Issue #7, opened 2015-11-02
[Feature Request] Minimize to tray


Sigmanor:
Hi. Can you add feature for minimize windows app to tray.

antelle:
Thanks, I'll look into it today.

antelle:
Fixed in 7df0328e1dfce93f41f70d0b59ecb4dfa9625fc9

antelle:
merged to master

PR #8, opened 2015-11-02
Fixed Minimize to tray #7

renatn:
Please review!

### DIFF
--- a/electron/main.js
+++ b/electron/main.js
@@ -6,11 +6,16 @@
 var app = require('app'),
     BrowserWindow = require('browser-window'),
     path = require('path'),
-    fs = require('fs');
+    fs = require('fs'),
+    Tray = require('tray'),
+    Menu = require('menu');
 
 var mainWindow = null,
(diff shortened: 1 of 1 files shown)


4ndv:
Needs option for changing behavior between minimizing and closing app, i think.

antelle:
Such default behaviour will bother users. I agree with @4ndv, this should be optional (but for now, there's no infrastructure to save options to main process). I'll add options engine and then merge this as a checkbox in settings, thank you!

4ndv:
Should auto-logout when minimizing to tray.

antelle:
So, I've implemented saving settings from webpage to app; now, the only blocker is locking (ux and flow): for now, only close is implemented.

renatn:
:thumbsup:

antelle:
So far, it has been merged to develop in 7df0328e1dfce93f41f70d0b59ecb4dfa9625fc9.
Changes I've made:
- lock on minimize
- minimize only on Windows (on Mac OSX, it's uncommon)
- checkbox in app settings

Will be available in upcoming v0.4.
Thanks for contribution!

PR #42, opened 2015-11-19
Minor Interface changes

Gutem:
Add titles to some icons.
Change "return to app" to Close icon.


antelle:
Thanks for contribution!
Titles are useful, I'll remove obvious ones and merge them now.
I'm going to work with titles in help/tips task. It also involves custom tooltips creation, so they will be not so ugly.

I disagree with close button design, I think what to do with it. Now it looks:
1. in wrong position, a bit lower than expected
2. too heavy and bright cross
So I'll roll it back.

antelle:
Merged into develop, thanks

PR #91, opened 2016-02-18
Add possibility to load file from url

ndusart:
Hello,

Here's a feature request in the form of a pull request.
I suspect that you wouldn't accept as is but I'm open to modify it to respect your conventions I could have missed.

The feature is to be able to load a keepass file directly from an url and let the app download it.
It works perfectly for uris in the same domains that the one hosting the keeweb app.
For other domains, the header 'Access-Control-Allow-Origin' should be set.

It can be convenient if you store your file on a web server. It saves the user to first download the file, load it in the app, and optionally not forgetting to delete it afterwards.

Thanks for the great app by the way :)

### DIFF
--- a/app/scripts/comp/alerts.js
+++ b/app/scripts/comp/alerts.js
@@ -16,16 +16,16 @@ var Alerts = {
         Alerts.alertDisplayed = true;
         var view = new ModalView({ model: config });
         view.render();
-        view.on('result', function(res, check) {
+        view.on('result', function(res, check, textinput) {
             Alerts.alertDisplayed = false;
             if (res && config.success) {
-                config.success(res, check);
+                config.success(res, check, textinput);
             }
             if (!res && config.cancel) {
                 config.cancel();
             }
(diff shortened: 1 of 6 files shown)


ndusart:
Sorry, I had branched from your master, I synced my master branch with your develop branch to have a tidier pull request :)

antelle:
Hello!
URL sync will be implemented as part of WebDAV protocol and configured as a storage provider... For now, I'm working on providers in a separate branch (I've not pushed it yet), they will be loaded as modules and plugged into the app. Webdav will be one of them, with its settings and name.
There will be a button on start screen to turn on different providers, so the UI will be a bit more complex than just a field on homepage.

ndusart:
Will it support web servers which do not support WebDav specific methods ?

antelle:
Yes, I think

ndusart:
Great ! Thanks and good work then.
Looking forward to see that :-)

PR #128, opened 2016-02-28
Urlcopy keyboardshortcut

baumhoto:
added keyboard shortcut to copy url to clipboard (Ctrl/Cmd+u)

PR #129, opened 2016-02-28
Support for electron global keyboard shortcuts

baumhoto:
this one is based on #128 and additionally adds global shortcut keys to electron-app, so copying different attributes to clipboard (user, password, url) doesn't require to switch back to app once the entry is selected.

Global shortcut keys added for copying attributes to clipboard: password(c), user(u), url(u)
OSX: Ctrl+Alt+(c/b/u) 
win/linux: Shift+Alt+(c/b/u)

Don't know if that is a feature you want to support?

### DIFF
--- a/electron/app.js
+++ b/electron/app.js
@@ -8,7 +8,8 @@ var app = require('app'),
     fs = require('fs'),
     BrowserWindow = require('browser-window'),
     Menu = require('menu'),
-    Tray = require('tray');
+    Tray = require('tray'),
+    globalShortcut = require('electron').globalShortcut;
 
 var mainWindow = null,
     appIcon = null,
@@ -29,6 +30,7 @@ process.argv.forEach(function(arg) {
 app.on('window-all-closed', function() {
     if (restartPending) {
         // unbind all handlers, load new app.js module and pass control to it
+        globalShortcut.unregisterAll();
         app.removeAllListeners('window-all-closed');
         app.removeAllListeners('ready');
         app.removeAllListeners('open-file');
@@ -44,6 +46,9 @@ app.on('window-all-closed', function() {
     }
 });
 app.on('ready', function() {
+    createGlobalShortcut('Ctrl+Alt+B', 'copy-user');
+    createGlobalShortcut('Ctrl+Alt+C', 'copy-password');
+    createGlobalShortcut('Ctrl+Alt+U', 'copy-url');
     createMainWindow();
 });
 app.on('open-file', function(e, path) {
@@ -57,6 +62,9 @@ app.on('activate', function() {
             createMainWindow();
         }
     }
+});
+app.on('will-quit', function() {
+  globalShortcut.unregisterAll();
 });
 app.restartApp = function() {
     restartPending = true;
@@ -234,4 +242,15 @@ function notifyOpenFile() {
             ' else { window.launcherOpenedFile="' + openFile + '"; }');
         openFile = null;
     }
+}
+
+function createGlobalShortcut(shortcut, backboneEventname)
+{
+    var ret = globalShortcut.register(shortcut, function() {
+        emitBackboneEvent(backboneEventname);     
+    });
+
+  if (!ret) {
+    console.log('registration failed');
+  }
 }
--- a/electron/app.js
+++ b/electron/app.js
@@ -46,9 +46,17 @@ app.on('window-all-closed', function() {
     }
 });
 app.on('ready', function() {
-    createGlobalShortcut('Ctrl+Alt+B', 'copy-user');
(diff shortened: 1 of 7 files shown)


antelle:
I've closed the previous PR as this includes it.
So, to merge it, please fix the following:
- join logic from [copyUserKeyPress](https://github.com/antelle/keeweb/blob/develop/app/scripts/views/details/details-view.js#L299) and copyUrlKeyPress to common method
- add shortcuts description to [settings/shortcuts](https://github.com/antelle/keeweb/blob/develop/app/templates/settings/settings-shortcuts.hbs#L10) page with corresponding [locale key](https://github.com/antelle/keeweb/blob/develop/app/scripts/util/locale.js#L291) - system-wide shortcuts should be the last
- style and spacing comments, I'll add them to the code

electron/app.js
```diff
@@ -44,6 +46,17 @@ app.on('window-all-closed', function() {
     }
 });
 app.on('ready', function() {
+    if (process.platform === 'darwin') {
+        createGlobalShortcut('Ctrl+Alt+B', 'copy-user');
+        createGlobalShortcut('Ctrl+Alt+C', 'copy-password');
+        createGlobalShortcut('Ctrl+Alt+U', 'copy-url');
+    }
```
antelle: `} else {`

electron/app.js
```diff
@@ -58,6 +71,9 @@ app.on('activate', function() {
         }
     }
 });
+app.on('will-quit', function() {
+  globalShortcut.unregisterAll();
```
antelle: 4 spaces

electron/app.js
```diff
@@ -235,3 +251,14 @@ function notifyOpenFile() {
         openFile = null;
     }
 }
+
+function createGlobalShortcut(shortcut, backboneEventname)
```
antelle: ) {

electron/app.js
```diff
@@ -235,3 +251,14 @@ function notifyOpenFile() {
...
+{
```
antelle: 4 spaces

electron/app.js
```diff
@@ -235,3 +251,14 @@ function notifyOpenFile() {
...
+    var ret = globalShortcut.register(shortcut, function() {
+        emitBackboneEvent(backboneEventname);     
+    });
+
+  if (!ret) {
+    console.log('registration failed');
```
antelle: just remove it please

electron/app.js
```diff
@@ -44,6 +46,17 @@ app.on('window-all-closed', function() {
...
+    if (process.platform === 'darwin') {
+        createGlobalShortcut('Ctrl+Alt+B', 'copy-user');
+        createGlobalShortcut('Ctrl+Alt+C', 'copy-password');
+        createGlobalShortcut('Ctrl+Alt+U', 'copy-url');
+    }
+    else
+    {
+        createGlobalShortcut('Shift+Alt+B', 'copy-user');
+        createGlobalShortcut('Shift+Alt+C', 'copy-password');
+        createGlobalShortcut('Shift+Alt+U', 'copy-url');
```
antelle: please make common event `copy-field` parametrized with the field itself

baumhoto: don't understand what i need to do here, so if you do it i'm fine with that:)

baumhoto:
i've updated the commit and fixed most of the issues.
- i actually refactored the 3 copyKeyPress methods into 1. hope your fine with that
- fixed code formatting issues etc.
- i modified the shortcuts-settings page, but there are still some issues:
  - it seems that this shortcut.css style does not support control key (at least not on the mac), or it does not support 2 modifier keys? it only display the Alt (Option) Key
  - in general this global modifier keys should only be displayed if user runs the electron app, but not if he uses the website, but don't know how i could do this in hbs file
 - currently  mac uses different global shortcut keys (control instead of shift), thus depending on the underlying platform the global shortcut keys differ. reason for using different modifier keys is that I'm used the mac combination (Ctrl+Alt) which does not work on windows. i have no problem to change it to Shift+Alt on mac for this pull request

antelle:
Okay, I'll fix this myself and merge this manually tonight. Actually, currently there's a bug with unsubscribe [here](https://github.com/antelle/keeweb/pull/129/files#diff-dc353f1b741bbe10d926b0e6eeee948cR72) which is not performed, as subscribe is made with an anonymous function and unsubscribe with named - I'll fix it as well.

baumhoto:
thanks, yes you're absolutely right i missed those calls in unsubscribe

antelle:
Merged it, thank you! Build with your changes deployed to https://beta.keeweb.info/

PR #139, opened 2016-03-01
import db from xml file

exromany:
It depend on [PR](https://github.com/antelle/kdbxweb/pull/6)

### DIFF
--- a/app/scripts/models/app-model.js
+++ b/app/scripts/models/app-model.js
@@ -258,6 +258,13 @@ var AppModel = Backbone.Model.extend({
         this.addFile(newFile);
     },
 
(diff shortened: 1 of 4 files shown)


antelle:
Thank you! :gift: 
I'll merge this into v1.1, after I finish storage providers UI. There will be a button `...` on welcome screen which will open a dialog with provider selection. This button will be among other options.
Having it on the very welcome screen is a bit weird.

antelle:
Merged in dccb1b2bf5220662d4af3fd3cbbdd382483d9a83
I've added some fixes for desktop and hidden button which will be revealed when we have storage options, under More button.
Thank you!

PR #145, opened 2016-03-02
Left and Right Hand Password Creation

steliosph:
Hi,

I have implemented the feature request for  https://github.com/antelle/keeweb/issues/142.
Please revise and let me know if there is something else.

There might be a better way to handle the Object.keys map function but at this moment i do not know one.

### DIFF
--- a/app/scripts/util/password-generator.js
+++ b/app/scripts/util/password-generator.js
@@ -4,11 +4,11 @@ var kdbxweb = require('kdbxweb');
 
 var PasswordGenerator = {
     charRanges: {
-        upper: 'ABCDEFGHJKLMNPQRSTUVWXYZ',
-        lower: 'abcdefghijkmnpqrstuvwxyz',
-        digits: '123456789',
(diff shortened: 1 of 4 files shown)


antelle:
Hello! Thanks for a contribution! 
But I'm afraid, it cannot be merged.
The feature you introduce here is a [delighter](https://upload.wikimedia.org/wikipedia/commons/6/68/Kano_model_showing_transition_over_time.png) one, not commonly used. So, having it on main screen will only confuse users (WTF is left hand / right hand??? - that would be the first reaction).
If you want to implement it, please follow this way:
- add a preset option 'one-hand type' to [this](https://github.com/antelle/keeweb/blob/develop/app/templates/generator.hbs#L5) presets list which will generate left- or right-hand passwords, randomly
- add a separate range to charRanges, please don't split existing ones, as this is very special option: tomorrow we may have down-top, and we don't want to split them to 3 dimensions
- don't hard-core strings: add them to locale.js and use with `{{res ...}}`
- make sure you work on develop branch, not master

PR #160, opened 2016-03-12
updating to add apple meta tags

morriswchris:
I've been using the hosted keeweb application on iOS and decided to save it to my home screen. Figured it would be a useful patch to remove the safari navigation bars, and make the home screen bookmark work like a standalone app.


antelle:
Hello!
You should've worked on `develop` branch, `apple-mobile-web-app-capable` is already there.
Regarding status bar color - isn't it black by default?

morriswchris:
Ahh my mistake. I guess the latest develop hasn't been merged down for the web hosted app. No need to merge this pull since it has already been completed.

PR #188, opened 2016-04-10
Textarea color as input color.  Fixes #187

### LINKED ISSUE #187: Textarea text color is always black, even in dark themes.
Steps to reproduce
- Open new or demo db.
- Select Flat blue theme in settings
- Go back to any entry, edit the Notes field, see how it's black text on dark.

Seems to me that it should be the same color as the input text is when editing.

I did read the banned features list, and if this is part of the "don't mess with my colors" banned feature, and not a bug, that's fine, happy to see this closed as "won't fix" and have this issue record it explicitly.  As a warning for the next person that wants to mess with the colors :smile: 

I reproduced it in Chrome windows, Firefox windows, and Chrome android.

textarea color

MrP:
Fixes #187 (Textarea text color is always black, even in dark themes.)

This seems like a bug to me, but I understand you have some opinions about the colors. 
Please do disregard this if it's not what you want.

Thanks anyway for your time,
Victor

### DIFF
--- a/app/styles/base/_forms.scss
+++ b/app/styles/base/_forms.scss
@@ -90,6 +90,7 @@ input[type=text], input[type=password], textarea, input:not([type]) {
 
 textarea {
   resize: vertical;
+  color: inherit;
 }
 
 input[type="search"] {


antelle:
Already fixed in cfcd54a1a3eae351b9a5ac6bdfcbf9f6ec9752d1

PR #233, opened 2016-05-21
fix #113: increased max password length

### LINKED ISSUE #113: Password being truncated?
When I create a new database and set a long password (1024+ chars), it won't work.

Steps to reproduce:
- create new db
- set master password to a long password
- (no key file set)
- sync with dropbox (did not test locally)
- logout
- try to login with same password
- login fails

I tried 128, 1024 and 2048-characters long passwords. The 128 one worked fine but both 1024 and 2048 failed. Anyone else can confirm?

KeeWeb v1.0.4 (a6e7afe, 2016-02-25)
Environment: web
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/48.0.2564.116 Safari/537.36

smacleod:
0198828ca74d0eb81c24c6887717796ecd06ff56 missed a couple places to increase the max password length


antelle:
👍

PR #244, opened 2016-06-15
Fix broken links to antelle/keeweb

grinsted:
fixed some broken links in the readme.


antelle:
👍 
it's already fixed in `develop` but old links were still here in `master` 
thanks!

PR #251, opened 2016-06-28
Wordage change

andyklimczak:
In password generator, change 'medium length' -> 'medium'.

This makes it consistant between 'medium' and 'long', and I thought just 'medium' and 'long' looked better than 'medium length' and 'long length'. (Fairly clear from the context)


antelle:
Thanks for pointing that out: it sounds really strange.
But just _medium_ doesn't work here: medium strength? security? It's confusing.
Another point is that its length is equal to default preset length, so long and longest won't work as well.
I'll think how to rename it.

andyklimczak:
I see your point. Just medium can mean different things unlike long, which refers to only one thing, the length.

I'll close this then, never mind !

PR #281, opened 2016-07-21
Fix #279: Auto-type with special characters on Mac

### LINKED ISSUE #279: Auto-type with special characters not working on Mac
Using Mac OS X 10.11.5.

I verified how the code creates an Applescript and spanws the execution of `osascript` with it. From what I checked, I got different results depending on the Input Source enabled on my OS.

This is a test script I created:
```
tell application "System Events"
keystroke "¼"
keystroke "Î"
keystroke "©"
keystroke "~"
keystroke "{"
keystroke "*"
keystroke "!"
keystroke "+"
keystroke "¢"
keystroke "f"
keystroke "4"
end tell
```

This is the result I got by running it having three different Input Sources selected:
* U.S.: `aÎ©~{*!+¢f4`
* Spanish - ISO: `aa©~{*!+¢f4`
* Brazilian - ABNT2: `¼a©~{*!+¢f4`

Expected: `¼Î©~{*!+¢f4`. Actually none of them got it right.  :- (

Plus, we have to consider that the Input Source affects how an accented character is typed. In U.S. we would type `Î` as <kbd>Alt</kbd>+<kbd>i</kbd>, <kbd>I</kbd>. In Spanish - ISO or Brazilian - ABNT2 we would type it as <kbd>^</kbd>, <kbd>I</kbd>. Maybe if we avoid using `keystroke`...

ericbn:
- Using copy-and-paste for special characters so it works on any Mac
  Input Source
- See http://stackoverflow.com/q/38484431/2654518
- Current delay of total 1.5 seconds per copy-and-paste makes
   auto-type considerably slow
- It could be faster if there was no splitting of characters in
  auto-type-runner.js getEntryFieldsKeys() when value.isProtected

### DIFF
--- a/app/scripts/auto-type/emitter/auto-type-emitter-darwin.js
+++ b/app/scripts/auto-type/emitter/auto-type-emitter-darwin.js
@@ -65,9 +65,11 @@ AutoTypeEmitter.prototype.key = function(key) {
 };
 
 AutoTypeEmitter.prototype.copyPaste = function(text) {
+    this.pendingScript.push('delay 0.5');
     this.pendingScript.push('set the clipboard to "' + this.escape(text) + '"');
+    this.pendingScript.push('delay 0.5');
     this.pendingScript.push('keystroke "v" using command down');
-    this.pendingScript.push('delay 0.05');
+    this.pendingScript.push('delay 0.5');
     this.callback();
 };
 
--- a/app/scripts/auto-type/emitter/auto-type-emitter-darwin.js
+++ b/app/scripts/auto-type/emitter/auto-type-emitter-darwin.js
@@ -36,10 +36,21 @@ AutoTypeEmitter.prototype.setMod = function(mod, enabled) {
     }
 };
 
+AutoTypeEmitter.prototype.isAlphaNumeric = function(text) {
+    return !(/[^a-zA-Z0-9]/.test(text));
+};
+
+AutoTypeEmitter.prototype.escape = function(text) {
+    return text.replace(/"/g, '\\"').replace(/\\/g, '\\\\');
+};
+
 AutoTypeEmitter.prototype.text = function(text) {
-    text = text.replace(/"/g, '\\"').replace(/\\/g, '\\\\');
-    this.pendingScript.push('keystroke "' + text + '"' + this.modString());
-    this.callback();
+    if (this.isAlphaNumeric(text)) {
+        this.pendingScript.push('keystroke "' + this.escape(text) + '"' + this.modString());
+        this.callback();
+    } else {
+        this.copyPaste(text);
+    }
 };
 
 AutoTypeEmitter.prototype.key = function(key) {
@@ -54,11 +65,9 @@ AutoTypeEmitter.prototype.key = function(key) {
 };
 
 AutoTypeEmitter.prototype.copyPaste = function(text) {
-    this.pendingScript.push('delay 0.5');
-    this.pendingScript.push('set the clipboard to "' + text.replace(/"/g, '\\"') + '"');
-    this.pendingScript.push('delay 0.5');
+    this.pendingScript.push('set the clipboard to "' + this.escape(text) + '"');
     this.pendingScript.push('keystroke "v" using command down');
-    this.pendingScript.push('delay 0.5');
+    this.pendingScript.push('delay 0.05');
     this.callback();
 };
 


antelle:
Thanks 🎁 
The reason why I set the delay to 0.5 was a bug while using auto-type obfuscation... I'm thinking how to fix this. Most probably now it became broken. I'll check it before merge.

ericbn:
You're, we have to keep it like it was. It can get very slow now, since it will copy-and-paste each non-alpha-numeric character separately in the password. So a password with 10 of them, for example, can get 15 seconds to get auto-typed.

It could be improved if there was no splitting of characters in auto-type-runner.js getEntryFieldsKeys() when value.isProtected. What is the reason behind this? Splitting the password in separate keystrokes does not prevent keyloggers from fetching it.

antelle:
Yes, it would be unacceptably slow, that's why I haven't just merged this fix, changing timeout to 0.5  😉 
I want to keep the timeout = 0.5 (better to fix it without timeout though but it seems to be impossible) in copy-paste action and insert 0.05 in characters from keystroke. Will this work? I'm not sure for now, we have to test it, with different combinations, e.g. subsequent `aâaâaâaâaâaâ`, or `aâa{left}{right}âa` and so on.

ericbn:
Actualy I haven't tested it enough before pushing the first commit. We'll need the same delay for both the new logic I'm introducing and what was using copyPaste() before.

ericbn:
Actually, a good and simple password to test with is `^A~A´A`.

Why are we spllitting the password in auto-type-runner.js getEntryFieldsKeys() when value.isProtected? If password was just a single string (as is the username), then auto-type would either paste the whole string or type the whole string in one command (depending on the presence or not of non-alpha-numeric characters, respectivelly).

Splitting the password in separate keystrokes does not prevent keyloggers from fetching it. Actually, neither adding random keystrokes is secure, since one can figure out what is the constant part by comparing a few keylogs of the same auto-typed password. The original KeePass has a good obfustation technique that mixes keystrokes with pasted content. See http://keepass.info/help/v2/autotype_obfuscation.html

antelle:
> Why are we splitting the password in auto-type-runner.js

Actually, we are not splitting it. We don't store passwords in memory, they are stored as xored chars, so if one inspects the app memory or app crashdump with memory, there're no passwords in it. This logic is implemented in ProtectedValue. That's why we don't join strings from ProtectedValues, we don't want the passwords to be stored in memory for indefinite time. Later we'll also have better input controls which will not require toString'ing ProtectedValues as well. So, joining passwords is not an option.

Also, obfuscation can make a long sequence of keystrokes and copypasted keys. Here's an example of emitted commands generated by obfuscating string `johnsmith` which we can deal with:
```
[auto-type-emitter] text e
[auto-type-emitter] key left
[auto-type-emitter] text i
[auto-type-emitter] key right
[auto-type-emitter] mod + true
[auto-type-emitter] key left
[auto-type-emitter] mod + false
[auto-type-emitter] text Q
[auto-type-emitter] key left
[auto-type-emitter] key bs
[auto-type-emitter] text s
[auto-type-emitter] key left
[auto-type-emitter] text 1
[auto-type-emitter] key del
[auto-type-emitter] text c
[auto-type-emitter] key left
[auto-type-emitter] text A
[auto-type-emitter] key left
[auto-type-emitter] text x
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] mod + true
[auto-type-emitter] key right
[auto-type-emitter] mod + false
[auto-type-emitter] text v
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key del
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] text Z
[auto-type-emitter] key del
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] text j
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key del
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] text t
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] copyPaste n
[auto-type-emitter] text 3
[auto-type-emitter] key left
[auto-type-emitter] copyPaste m
[auto-type-emitter] key bs
[auto-type-emitter] text t
[auto-type-emitter] key left
[auto-type-emitter] text m
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] text j
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] text h
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] key left
[auto-type-emitter] text i
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key bs
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] key left
[auto-type-emitter] text o
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] text s
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] key right
[auto-type-emitter] text h
[auto-type-emitter] waitComplete
```
So, yes, I have to test it, and there are several options to fix:
- fix the delay issue and understand how to force Cocoa wait until the app has processed sent copy-paste event, like sync SendMessage, in contrast to PostMessage, in WinAPI. At the moment, I don't know how to do this, so I inserted that ugly delay. This is the best option but also the hardest or even impossible;
- buffer non-ascii keystrokes and paste them together when either (a) ascii key is added, (b) end of input is reached. This will still slow down obfuscated input but we can make a sort of warning that non-ascii characters input may be slow;
- maybe there's a way to add them without using copy-paste? I'm not sure...

ericbn:
I found this: https://github.com/keithbennett/keepassx/blob/master/src/autotype/mac/AutoTypeMac.cpp

Tested with `^A~A´A¼Î©~{*!+¢f4` and it worked beautifully. Can we include C++ code with Electron?

antelle:
It's using Carbon events. Well, if all else fails, we can create a native app helper, like in Windows. Most probably we'll have to do it. Another implementation: http://stackoverflow.com/questions/10734349/simulate-keypress-for-system-wide-hotkeys/27484122#27484122

antelle:
So, I created a simple test helper app with swift, here's the draft: https://github.com/keeweb/keeweb/blob/develop/helper/darwin/KeeWebHelper/KeeWebHelper/main.swift
It works for me if en and ru locales. So, most probably, in v1.2 we'll use a helper for mac as well.

ericbn:
Cool! Can't we use [Node.js Addons](https://nodejs.org/api/addons.html) instead of an external executable? I can help with the C/C++ code...

antelle:
Addons require more complex support because they cannot be unloaded and updated without app restart. No, currently we are not planning to use addons, this will be an executable.

antelle:
The helper [was implemented](https://github.com/keeweb/keeweb/tree/develop/helper/darwin) and now I have to test it, then integrate with keeweb. Basic cases in different layouts, input sources and character cases are working.

antelle:
closing this, now it's not actual

PR #297, opened 2016-08-04
Fix favicons and images from being stretched horizontally

mikfoo:
When downloading favicons or setting custom icons, they were 2px too wide.

**Before:**
before

**Now:**
after


antelle:
🎁

PR #332, opened 2016-08-28
One time password with spaces

izissise:
Some websites (google) give you the key separated with spaces
This try to remove spaces and see if it is a valid Otp.


antelle:
🎁  thank you

PR #438, opened 2016-11-13
Present all obfuscated passwords with the same length.

RhysBower:
This would prevent someone looking over your shoulder and guessing the length of your password. A slight improvement in security through obscurity for the the uninitiated hacker.

### DIFF
--- a/app/scripts/util/password-generator.js
+++ b/app/scripts/util/password-generator.js
@@ -110,7 +110,7 @@ var PasswordGenerator = {
     },
 
     present: function(length) {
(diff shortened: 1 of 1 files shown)


antelle:
- this must be done in presentation (view), not in helper class, because `present` method accepting length must not return constant.
- it's used in several places, we don't want to have it everywhere
- empty passwords must be displayed as empty
- sometimes, it's useful to see the password, it would be better to masking as a setting
- globally recognisable symbol is three bullets, without spaces

PR #439, opened 2016-11-13
Fixed WebDav save destination

### LINKED ISSUE #435: WebDav incorrect move destinations
(very similar to #205)

When saving via WebDav, a HTTP 400 error is returned if the webdav path is relative (eg `webdav/<name>.kdbx`, and the URL used to load keeweb contains an additional slash (I'm using `/keeweb/?config=/config.json`). A 405 error is generated if the WebDav path is relative to the root (eg `/webdav/<name>.kdbx`, and the directory from which KeeWeb is being served is not the root.

Since the replacement is done on the final slash in the URL, the move destination becomes `https://domain/keeweb/?config=/webdav/<name>.kdbx`, rather than just `https://domain/webdav/<name>.kdbx`.

I also noticed that if using a path relative to the root (eg `/webdav/...`), the current replacement results in a incorrect path being generated, also including a double slash, since the webdav path is simply appended onto the existing path. Eg, the above example becomes `https://domain/keeweb//webdav/<name>.kdbx`. If the specified webdav path starts with a /, I guess that the entire URL could be built up from the individual components (eg `location.protocol + "//" + location.host`)

TimStallard:
Modified WebDav save move destination to work correctly with relative and root paths, removing any URL parameters to prevent replacing the wrong /, fixes #435

### DIFF
--- a/app/scripts/storage/storage-webdav.js
+++ b/app/scripts/storage/storage-webdav.js
@@ -86,7 +86,11 @@ var StorageWebDav = StorageBase.extend({
                     }
                     var movePath = path;
                     if (movePath.indexOf('://') < 0) {
-                        movePath = location.href.replace(/[^/]*$/, movePath);
+                        if (movePath.indexOf('/') === 0) {
+                            movePath = location.protocol + '//' + location.host + movePath;
+                        } else {
+                            movePath = location.href.replace(/\?(.*)/, '').replace(/[^/]*$/, movePath);
+                        }
                     }
                     that._request(_.defaults({
                         op: 'Save:move', method: 'MOVE', path: tmpPath, nostat: true,


antelle:
👍 
Hope it works (have no ability to test it right now, but looks legit, thank you!

TimStallard:
I tested it with full, relative and root paths, all of which worked.

I was only able to test the web version, but this code should never be called on the desktop versions anyway, since http(s):// would have to be included in the webdav path for those.

TimStallard:
BTW, how regularly is the gh-pages branch updated - when can I expect this change to be available there?

antelle:
It's updated when release happens, not too often. The next one will be approx. in a month, I think. [Beta](https://beta.keeweb.info/) is updated more often from `develop` branch, but I have no possibility to build the app now, so the best way to get it now is build it yourself...

PR #463, opened 2016-12-11
A setting to remember only the path of key file (desktop apps)

arantir:
Reasoning: 
A key file if often used for 2FA, being stored e.g. on a USB stick, which you (try to) keep separately. It's hard (or impossible) to provide a reasonable security (like encryption) to all of the devices you use, so you keep the key away unless it's needed, in case a device is stolen or hacked.
This PR provides a feature to remember the path of last key file used, so you don't have to select every time. You can just put the USB stick, type the password and you're in.

Feature notes:
* The new setting option is only available to desktop apps.
* On the homecreen (open view) you will see a full path if just the key file path is stored and the filename if the key itself is stored.

Implementation notes:
* I clear the key file data when the setting is changed (because whichever option is chosen we would want to forget the data related to the previous option).
* I store the key file path only if the setting is checked, thought it makes the code simpler in some cases.

**Help or suggestions needed:**
* The setting lable translations to other embedded locales. I guess Google Translate is a bad fellow for this.


app/scripts/views/settings/settings-general-view.js
```diff
@@ -235,8 +238,20 @@ var SettingsGeneralView = Backbone.View.extend({
     changeRememberKeyFiles: function(e) {
         var rememberKeyFiles = e.target.checked || false;
         AppSettingsModel.instance.set('rememberKeyFiles', rememberKeyFiles);
-        if (!rememberKeyFiles) {
-            this.appModel.clearStoredKeyFiles();
+        this.appModel.clearStoredKeyFiles();
```
antelle: should be in 'else' block

antelle:
Hello,
This looks great 👍 
Have you tested it with Dropbox, does it work ok?
Please also replace two checkboxes with a select control. Everything else looks ok.

arantir:
Dropbox-based key file selection is available only in browser, so it's a bit different case. The new option doesn't change (neither breaks) how Dropbox key file selection works in browser. 
I'm almost sure for those who store the key file in Dropbox it is much less of a consern to not cache the key file. This may be implemented separately if requested.

Basically, the option is for the file storage only. For desktop apps it works just seamlessly with filesystem-integrated storages (checked on Google Drive, but would work with anything which allows the file to opened by a path). 

Update notes:
* Still always clear the key file data when the setting is changed (because whichever option is chosen we would want to forget the data related to the previous option).
* Changed the bahaviour of the key file selection field on the homescreen to be more consistent and intuitive. 
Now, when clicked, it will always forget the value if there's any, and will always open the file selection dialog if empty.
Previously it would clear the value only if there's key file data, which is not jet loaded when you close a database and try to open it again, in both cases (stored key/stored path), so it behaved differently before and after you open a database.

PR #530, opened 2017-02-16
[macOS] Title bar style

edipox:
Already tested in Linux, macOS and browser. It works perfectly because it only affects the macOS app.

keeweb_new
keeweb

### DIFF
--- a/app/templates/menu/menu.hbs
+++ b/app/templates/menu/menu.hbs
@@ -1,2 +1,2 @@
-<div class="menu">
+<div class="menu {{platform}}">
 </div>
--- a/app/scripts/helpers/index.js
+++ b/app/scripts/helpers/index.js
@@ -5,3 +5,4 @@ require('./ifeq');
 require('./ifneq');
 require('./ifemptyoreq');
 require('./res');
+require('./platform');
--- a/app/scripts/helpers/platform.js
+++ b/app/scripts/helpers/platform.js
@@ -1,0 +1,7 @@
+'use strict';
+
+const Handlebars = require('hbs');
+
+Handlebars.registerHelper('platform', function() { // eslint-disable-line prefer-arrow-callback
+    return process.platform;
+});
--- a/app/styles/areas/_menu.scss
+++ b/app/styles/areas/_menu.scss
@@ -6,6 +6,10 @@
   @include justify-content(flex-start);
   overflow: hidden;
 
+  &.darwin{
+    margin-top: 22px;
+  }
+
   &__section {
     @include flex(0 0 auto);
     @include th { border-bottom: light-border(); }
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -113,6 +113,7 @@ function createMainWindow() {
         show: false,
         width: 1000, height: 700, minWidth: 700, minHeight: 400,
         icon: path.join(__dirname, 'icon.png'),
+        titleBarStyle: 'hidden',
         webPreferences: {
             backgroundThrottling: false
         }


app/templates/menu/menu.hbs
```diff
@@ -1,2 +1,2 @@
-<div class="menu">
-</div>
\ No newline at end of file
+<div class="menu {{platform}}">
```
antelle: Please pass it in app-view, here: https://github.com/keeweb/keeweb/blob/develop/app/scripts/views/app-view.js#L104
We don't need a helper, to use it only one time.

antelle:
Cool, thanks for that! 👍 
Please make it as an option. It might disturb somebody, personally I like mac header more.

edipox:
Ok, thanks @antelle I'll take a look at how to make it as an option.

antelle:
It should be somewhere here: https://github.com/keeweb/keeweb/blob/master/app/scripts/views/settings/settings-general-view.js#L84
, and a checkbox on settings page.
In Atom it's called 'Use custom title-bar', you can put this name to locales/base.json as well.
Then read app-settings.json inside app.js and apply it.

EnricoMonese:
I'm implementing the setting as a dropdown from which you can choose:
- default
- hidden
- hidden-inset
These are the options given from electron [here](https://github.com/electron/electron/blob/master/docs/api/browser-window.md)

I then intend to pass it here https://github.com/keeweb/keeweb/blob/develop/desktop/app.js#L114, what would be the best way to get the value from _app-settings.json_ from there?

antelle:
@EnricoMonese just read app-settings.json there.
I guess, there's no possibility to change it in a running app, right? Then it's easier, you can just write in a label, that it will be applied only after restart.

EnricoMonese:
Yes it can only be set when the window is created.

antelle:
Cool, let's just mention it in the label then.

EnricoMonese:
By reading it do you mean open the file using fs and then parsing it, or is there already something that can be imported that wraps this funcionality?

antelle:
@EnricoMonese yes, just like window position:
https://github.com/keeweb/keeweb/blob/develop/desktop/app.js#L16
https://github.com/keeweb/keeweb/blob/develop/desktop/app.js#L208
But in your case, as we need this before we can create a window, you can just load it sync and parse.

EnricoMonese:
Did something fast but it throws some errors. Will take a better look and commit later as now I have to go. Probably just forgot a comma somewhere though. I also saw that there is something like sheduleRestart could that be used after changing the setting?

antelle:
Yes, it's used during updates, to trigger auto-update, you can use `Launcher.requestRestart();`.
If you want, you can add a button 'Restart me', which will appear after you change the mode and call this code from it.

EnricoMonese:
Should I open a new pr?

antelle:
@EnricoMonese, yes, I guess, you cannot modify this PR, the only way is to create a new one.

EnricoMonese:
Also having the different titlebar means that there no longer is a draggable region. I made the whole window draggable from everywhere as explained [here](https://github.com/electron/electron/blob/master/docs/api/frameless-window.md#draggable-region), but I guess that maybe that should be an option too

antelle:
@EnricoMonese hm, I can drag Atom in Custom Titlebar mode, I think it makes sense to make title region draggable, otherwise it's not usable at all.

EnricoMonese:
Will add a checkbox to make the whole window draggable, if set to false it will use either the default titlebar or just the top of the content when the titlebar is hidden.

antelle:
No, we don't need two settings just for this simple switch. Let's just make the title draggable? Are there any apps which are draggable by the whole window?

EnricoMonese:
There are some but it seems a common behavior in apps with the hidden inset titlebar. As I already did the change I can say that it actually isn't too distracting. It just means that you can also drag the window by clicking on blank areas. I personally find it useful in other apps so I just thought that since it has to be addressed anyway we could simply add a toggle.

antelle:
Let's add a draggable option to the bottom of the select.

antelle:
Moved to #571

PR #571, opened 2017-03-26
macOS titlebar style settings

EnricoMonese:
schermata 2017-03-26 alle 16 24 51

Settings to select titlebar style on macOS
a
- First one is default macOS
- Custom titlebar, top 30px act as draggable region
- Custom titlebar with inset _traffic light_ buttons, window is draggable from everywhere

When entering fullscreen the layout goes back to normal and fills the space of the window buttons.


EnricoMonese:
What would be the best way to make this setting only appear on macOS?

antelle:
Add an option to FeatureDetector: supporsTitleBarStyles, check for macos there.
Probably we could implement this for Windows later as well.

antelle:
Apart from this, looks cool, I'll prettify settings in electron app reading and merge it.
Thank you! 🎁

app/scripts/views/settings/settings-general-view.js
```diff
@@ -82,7 +82,10 @@ const SettingsGeneralView = Backbone.View.extend({
         const updateFound = UpdateModel.instance.get('updateStatus') === 'found';
         const updateManual = UpdateModel.instance.get('updateManual');
         const storageProviders = this.getStorageProviders();
+        const platform = (typeof process !== 'undefined') ? process.platform : 'web';
+
```
antelle: please don't do this

app/templates/settings/settings-general.hbs
```diff
@@ -64,6 +64,7 @@
             <option value="2" {{#ifeq fontSize 2}}selected{{/ifeq}}>{{res 'setGenFontSizeLargest'}}</option>
         </select>
     </div>
+    {{#ifeq platform 'darwin'}}
```
antelle: should be in feature-detector.js

app/scripts/views/settings/settings-general-view.js
EnricoMonese: Just saw your suggestion about the FeatureDetector 👍

antelle: :+1: cool

EnricoMonese:
Added it to `FeatureDetector: supporsTitleBarStyles` but it doesn't seem to work, what did I miss?

antelle:
You added it as a function, to it should be: supportsTitleBarStyles()

EnricoMonese:
Ah silly me, fixed it!

EnricoMonese:
Just realized that I made the pr for the master branch, should it be changed to the dev branch?

antelle:
yes, it's better to check on develop: some things have been changed

Issue #497, opened 2017-01-24
"Auto-lock on minimize" should probably also lock on hide (MacOS)


DarwinAwardWinner:
Thanks for taking the time to contribute!

I noticed that when the "Auto-lock on minimize" preference is enabled, hiding the app in Mac OS (i.e. pressing Command+H) doesn't cause the database to be locked. I'm not sure if this is intentional, but it seems to run counter to the spirit of this preference, if not the letter. I would suggest that hiding should count as minimizing for the purposes of auto-locking the database.

antelle:
So, the change will be reverted, because it shouldn't lock on auto-hide, e.g. when the window is hidden by auto-type. For now, there's no simple way to distinguish between them, so we'll leave it for later.

DarwinAwardWinner:
One simple way to distinguish between manual window hiding and window hiding as part of an auto-type action could be timing. If auto-type never hides the window for more than, say, 10 seconds, then just wait a minimum of ten seconds after a hide before locking, and cancel the lock if the window is un-hidden before then.

antelle:
Sounds very hackish, probably there's a better way.

EnricoMonese:
What about listening directly for <kbd>Command</kbd>+<kbd>H</kbd> ?

DarwinAwardWinner:
That wouldn't work, since there are many other ways to hide an application.

EnricoMonese:
Yeah that's true

EnricoMonese:
Maybe something like @DarwinAwardWinner said. The hide event is used but autotype sets a flag to stop it. But if left in the background it gets locked by the _lock after some time_ feature

antelle:
Yes, something like this, but not this way as it was described. We have to remember if we called hide ourselves and lock if it was initiated by the user.

PR #575, opened 2017-03-28
"Auto-lock on minimize" also lock on hide 

EnricoMonese:
Issue #497 
"Auto-lock on minimize" should probably also lock on hide


antelle:
cool, thanks, no objections, merged 🎁

Issue #587, opened 2017-04-07
[Feature request] option to auto-lock database after auto-type


lawrencium:
Currently there are options to auto-lock the database after minimize or password copy. I think it'd be useful to add an option to auto-lock after an auto-type.

antelle:
Yes, definitely. And we need something like multi-select for it then.

PR #590, opened 2017-04-08
option to auto-lock database after an auto-type action

lawrencium:
Pull Request for https://github.com/keeweb/keeweb/issues/587

Pretty much followed the structure of locking on minimizing & copy password. 

The option defaults to false (auto-type does not trigger lock unless user explicitly checks the box).

### DIFF
--- a/app/scripts/auto-type/index.js
+++ b/app/scripts/auto-type/index.js
@@ -1,5 +1,3 @@
-'use strict';
-
 const Backbone = require('backbone');
 const AutoTypeParser = require('./auto-type-parser');
 const AutoTypeFilter = require('./auto-type-filter');
--- a/app/scripts/auto-type/index.js
+++ b/app/scripts/auto-type/index.js
@@ -62,6 +62,10 @@ const AutoType = {
                 });
             }
         });
+
+        if (AppSettingsModel.instance.get('lockOnAutoType')) {
+            Backbone.trigger('lock-workspace');
+        }
     },
 
     run(entry, callback) {
@@ -143,10 +147,6 @@ const AutoType = {
             setTimeout(callback, Timeouts.AutoTypeAfterHide);
         } else {
             callback();
-        }
-
-        if (AppSettingsModel.instance.get('lockOnAutoType')) {
-            Backbone.trigger('lock-workspace');
         }
     },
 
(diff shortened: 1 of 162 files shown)


app/scripts/auto-type/index.js
```diff
@@ -143,6 +144,10 @@ const AutoType = {
         } else {
             callback();
         }
+
+        if (AppSettingsModel.instance.get('lockOnAutoType')) {
```
antelle: Please move to `runAndHandleResult`

app/templates/settings/settings-general.hbs
```diff
@@ -147,6 +147,12 @@
         <label for="settings__general-lock-on-copy">{{res 'setGenLockCopy'}}</label>
     </div>
 
+    <div>
```
antelle: This must not be visible on web. Please wrap with `{{#if canAutoType}}` and use `AutoType.enabled` for it.

antelle:
I just tested this, and `auto-lock on minimize` locks the file for me. So, the suggestion is to rename the option to 'minimize or auto-type'. What happens for you, which OS are you using?

lawrencium:
I'm on Mac OS X. I played around w/ auto-lock-on minimize, and I found it kind of confusing to think about, primarily (I think) due to the changes brought about by #497. 

The `auto-lock on minimize` option locks on auto-type if the window the auto-type is performed on completely overlaps the KeeWeb app. However, if you don't overlap the KeeWeb app and the window you are auto-typing on, then KeeWeb will not lock because it won't consider itself hidden. So, it is possible to have `auto-lock on minimize` enabled, complete an auto-type, and not have the database lock itself afterwards.

I'm not a big fan of #497 because I think it overloads the definition of "minimize" and I think there should be a separate option to `auto-lock on hide`.

antelle:
Yes, probably it makes sense to revert that change, and then it becomes clear.

lawrencium:
So, after playing around w/ the electron framework and reading #497, I think that the implementation of #497 (done in #575) isn't correct. An event listener for 'hide' is triggered whenever the window _is not visible_, meaning that a 'hide' event is emitted whenever the app cannot be seen. This is different than the intent of #497, which is looking to bind a Mac-specific type of event to 'minimize'.

I attached a commit to revert the changes in #575. I think the best way to handle #497 is to submit a pull request to the electron repo to either 1) emit a new event when a hide-action is performed or 2) also emit a 'minimize' event when a hide-action is performed.

Do you guys have thoughts about this?

antelle:
I absolutely agree, we can't treat automatic hide as a user action, so it's better to revert that commit.

lawrencium:
Thanks, so do you think this pull request can be merged?

antelle:
yes, of course, I'll merge it tonight

antelle:
Merged to develop, uploaded to beta

PR #591, opened 2017-04-08
KeeWeb helper not found

lawrencium:
I was getting an error when running the project through electron:

> Error Helper not found. Searched paths: ... (error stems from auto-type-native-helper.js)

I believe the issue is caused by the call to Launcher.statFileSync(possiblePath) on line 22. Looking at the method implementation (in launcher-electron.js, line 78), it returns undefined, yet AutoTypeNativeHelper tries to get a ctime property from it.


antelle:
Thanks a lot! 🎁 
It got broken during a refactoring...

PR #594, opened 2017-04-11
Adding cordova launcher

alex-shpak:
Hi.
This is not final version to merge, but I would like you to check it first.
Most things works, except clipboard, which requires to be async for cordova. Is there way to make it async with electron launcher?


app/scripts/comp/launcher-cordova.js
```diff
@@ -0,0 +1,216 @@
...
+                if (!result.isAvailable) {
+                    return;
+                }
+
+                const encryptConfig = _.extend({}, this.config, {
+                    username: fileInfo.id,
+                    password: password.getText()
+                });
+
+                FingerprintAuth.encrypt(encryptConfig, result => { // eslint-disable-line no-undef
```
antelle: Please add cordova globals to the beginning of the file:
```
/* global FingerprintAuth */
```

app/scripts/models/app-model.js
```diff
@@ -489,7 +490,7 @@ const AppModel = Backbone.Model.extend({
             }
             const rev = params.rev || fileInfo && fileInfo.get('rev');
             this.setFileOpts(file, params.opts);
-            this.addToLastOpenFiles(file, rev);
+            this.addToLastOpenFiles(file, fileInfo, params.password, rev);
```
antelle: Passwords in local storage? No, we cannot.

app/scripts/comp/launcher-electron.js
```diff
@@ -244,6 +244,9 @@ const Launcher = {
             }
         }
         return ps;
+    },
+    openFileChooser: function(callback, button) {
```
antelle: please don't access any UI elements in Launcher. Instead, you can create a property `canChooseFile` and check it in calling code.

app/scripts/storage/index.js
```diff
@@ -12,4 +12,11 @@ const ThirdPartyStorage = {
     onedrive: require('./storage-onedrive')
 };
 
-module.exports = _.extend({}, BuiltInStorage, ThirdPartyStorage);
+let storages;
+if (Launcher.name === 'cordova') {
```
antelle: Please don't check Launcher this way. Instead, create a property `canUseStorages` in Launcher and check it, but not here.
Remove this code to StorageBase, there's `enabled` property in it.

alex-shpak:
Hi.
In this implementation password is not stored in localStore (maybe I did something wrong?) but passed to cordova in clear text. In cordova password will be encrypted and stored using fingerprint API of smartphone, so clear text is only in communication between cordova and browser.
When user use fingerprint password will be decrypted and sent to browser.

Here: https://github.com/keeweb/keeweb/pull/594/files#diff-cbbbc85718b3b853f2931a17729b7b4cR554
I put it there as single point which called when file is added, maybe you can find better place.

There is fingerprint saved In local storage (token obtained from fingerprint API, which is not password)
Here: https://github.com/keeweb/keeweb/pull/594/files#diff-1a3bda52cc269d75018b897be609016fR193

As for other comments - will fix it soon :)

antelle:
So, what is passed here as `params.password`? 
https://github.com/keeweb/keeweb/pull/594/files#diff-cbbbc85718b3b853f2931a17729b7b4cR493

antelle:
Oh, I see, you just pass it there and don't store. Weird, it must be moved somewhere...

antelle:
Probably, the right place for it as the code which handles file opening and fingerprints management, i.e. open-view.

alex-shpak:
Sure, for fingerprint auth I needed `fileInfo` instance (to obtain `id` and save `token` later) and clear text password. Have some idea where we can put it?

antelle:
Can we move everything related to fingerprints into `OpenView.openDbComplete`?
Please also remove this call: https://github.com/keeweb/keeweb/pull/594/files#diff-1a3bda52cc269d75018b897be609016fR194
Launcher must not interact with app in active manner.

app/scripts/comp/launcher-cordova.js
```diff
@@ -0,0 +1,216 @@
...
+                    fileInfo.set('fingerprint', result.token);
+                    appModel.fileInfos.save();
```
antelle: should be removed

app/scripts/comp/launcher-cordova.js
```diff
@@ -0,0 +1,216 @@
...
+        window.cordova.exec(onFileSelected, callback, 'FileChooser', 'choose');
+    },
+
+    fingerprints: {
+        config: {
+            disableBackup: true,
+            clientId: 'keeweb'
+        },
+
+        register: function(appModel, fileInfo, password) {
```
antelle: This method should accept { id, password } and return a token.

alex-shpak: Just to note: method is async

antelle: Of course, I mean the conceptual return. I.e. for async method you should add a callback or promise.

app/scripts/comp/launcher-electron.js
alex-shpak: Is it okay to check that method exists in launcher? e.g
```
    if (Launcher && Launcher.openFileChooser) {
            Launcher.openFileChooser((err, file) => { });
        } else {
            fileInput.click();
        }
```

antelle: If the whole feature is represented as a single method, I think it's pretty ok to check it directly. :+1:

alex-shpak:
> Can we move everything related to saving password in fingerprint auth into OpenView.openDbComplete?

I checked it and not quite sure how to get all required info there.
I need fileInfo to get Id and safe token into it later, password and existing fileInfo (if file is not new)

P.S. Fixed other comments

antelle:
Cool thanks,
I'll merge it later and add do something with this fileinfo.

alex-shpak:
Thanks! I also would need few days to test it on my phone after changes.

antelle:
Thanks :gift: 
Merged to develop, added some changes, please test.
I commented out cordova launcher, since it's not complete for now.
Possible issues (I haven't tested this, just want to draw your attention):
- after password change it won't be updated
- files with passwords and keyfiles

alex-shpak:
Hi! Great.

Changing password I will need to add, `fingerprint` is stored for that in fileInfo
Files with keyfile I think should work since cordova just sets password to input field

Also in TODO is clipboard support

antelle:
Probably we also need an option to store fingerprints or not. I would not use the app if it stored passwords without a permission.
Also, I think, by removing Dropbox, Google, etc... sync options we're defeating one of the main advantages of using keeweb, so I would spend some time on it as well.

alex-shpak:
> Probably we also need an option to store fingerprints or not. I would not use the app if it stored passwords without a permission.
I think having option is good. Just to mention app cannot store password without user interaction (and user always can cancel it). Technically it possible, but that's bad design and I don't want it.
App also will ask for fingerprint permissions when installed.

> Also, I think, by removing Dropbox, Google, etc... sync options we're defeating one of the main advantages of using keeweb, so I would spend some time on it as well.
I don't know about iOS but in android Drive, Dropbox and others are available from Content Providers which is used by file chooser. That's why I removed them from UI. Also it might be hard to make OAuth flow in cordova I think.

antelle:
On iOS they are available as well, but the point to link file in keeweb is sync. What will happen if you select a file from chooser, make a change and save it, will it be uploaded back to Dropbox in Android? And vice versa, will the file be updated in app when it's updated in Dropbox?

antelle:
Here's how we can use Dropbox OAuth2 with Cordova: https://github.com/dropbox/dropbox-js/blob/stable/src/auth_driver/cordova.coffee
dropbox-js lib will be removed from keeweb soon, but we can implement something like this and it will work for all providers. It should be easy, since we're using own implementation of OAuth2. Well, I'll look into it later, if we manage to make Cordova working at all.

antelle:
Field editing is now more usable on mobile devices:
editing

alex-shpak:
Great!

I tried sync with Google Drive via content provider and it works, when I change file with desktop client and then open file on mobile (file was previously added) it get synced. Works in other direction as well.
I didn't try dropbox yet, but I will.

I will push what I have to github. Also I can provide you with built APK.

This is how it looks
image

Do you consider opening gitter chat for this repo? So then we can discuss things in more fluent way :)

antelle:
Hm, I think, conflict resolution doesn't work smoothly then. E.g. when you update a file on device being offline and sync with a new version later, it will be either replaced, or duplicated, instead of merge.
I've removed obsolete Dropbox lib, now we're using V2 endpoints and have the same oauth implementation for all storage providers, so it will be fairly easy to support Cordova, I'll look into it later and implement it, if I manage to launch the app on iOS. Should be as simple as overriding window.open and handling OAuth return.

Do you think we need any other changes for proper mobile support?
E.g. I pushed some fixes to field editor and footer, maybe something else behaves unnatural?
I can't really pay attention to chats, sorry, you can just email me.

alex-shpak:
Okay, I will check content provider behavior.
I also fixed some bugs related to mkdir in cordova launcher.

Things that comes to my mind for mobile:
 - Error handling needs to be reviewed, for example I seen handling error with `err.notFound` but cordova errors are different maybe we need mapping or something.
 - I need to update password saved with fingerprint when it changed in app or after sync with storage.
 - I think there is some serious changes required in interface (like side menu),
 - To try with iOS
 - Async clipboard operations (I didn't find sync way to do it in cordova)

Maybe something else will come to mind

PR #614, opened 2017-04-26
chore: remove IntelliJ dot files

timbru31:
This PR removes the `.idea` folder an `.iml` file belonging to the IntelliJ (WebStorm) project stuff.


antelle:
reason?

timbru31:
For people using other editors these files are just garbage files.  
People using the same editor might have issues when no variables are used for path values when importing the project.  
In general (IMO) files in a git repo should not contain editor settings from any author/contributor. For common formatting rules there are projects like ESLint or EditorConfig.

Up to you though ;)

antelle:
> For people using other editors these files are just garbage files.

Sorry, what? Ignore them, that's it.
> issues when no variables are used for path values when importing the project

Could you please provide an example of such errors? It should not happen. If it does, then it makes sense to ignore specific files, containing such paths.

antelle:
I just checked the latest official doc: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
All files from it are ignored, so it should not cause issues.

timbru31:
Well, I believe that the repository should be kept clean of any clutter, especially IDE or OS temp files. While you are right that these files might not harm, it's [common practice](https://github.com/spring-projects/spring-boot/blob/master/.gitignore#L30-L33) to ignore such files.
But I'm fine with your decision, as you already closed this PR. 🤷‍♂️

PR #674, opened 2017-06-26
Cordova specific file system and fingerprint fixes.

alex-shpak:


Thargarius:
This feature would be so awesome !!! :D

app/scripts/views/settings/settings-file-view.js
```diff
@@ -160,7 +160,16 @@ const SettingsFileView = Backbone.View.extend({
                 return;
             }
         }
-        this.appModel.syncFile(this.model, arg);
+
+        const password = this.$el.find('#settings__file-master-pass').val();
```
alex-shpak: This should be somehow properly refactored≥

alex-shpak:
@antelle Any chances for this PR? :)

antelle:
We can carefully merge it with the next release. Not sure if still makes sense (because of general concerns you brought up in another ticket), but yeah let's merge it.

app/scripts/views/settings/settings-file-view.js
antelle: I just wanted to merge this for the next release which should be coming soon, but noticed this. It's something that's better not to have in master.

app/scripts/models/app-model.js
```diff
@@ -645,7 +645,7 @@ const AppModel = Backbone.Model.extend({
             this.scheduleBackupFile(file, data);
         }
         if (params) {
-            this.saveFileFingerprint(file, params.password);
+            this.saveFileFingerprint(file, params.password.getText());
```
antelle: `getText` exposes the password as plaintext in memory. It should be called in the fingerprint handling code and passed from here as `password`.

antelle: In the fingerprint module in js, closer to the place where it’s used, should be enough. Now it will happen for everyone and too early.

alex-shpak: I will take a look

antelle: The place where I suggest moving them is `launcher-cordova.js::Launcher.fingerprints.register`.

alex-shpak: Done.

app/scripts/views/settings/settings-file-view.js
alex-shpak: I removed fingerprint update call since I didn't find another suitable workaround.
User will be asked to protect password with fingerprint on next login after password change.

app/scripts/views/settings/settings-file-view.js
```diff
@@ -163,7 +163,15 @@ const SettingsFileView = Backbone.View.extend({
                 return;
             }
         }
-        this.appModel.syncFile(this.model, arg);
+
+        // save to erase old password from fingerprint store after sync
+        const shouldUpdateFingerprint = this.model.get('passwordChanged') && this.model.get('fingerprint');
+
+        this.appModel.syncFile(this.model, arg, () => {
```
antelle: let's put it to `app-model::syncFile::complete`, before `save`.

app/scripts/views/settings/settings-file-view.js
```diff
@@ -163,7 +163,15 @@ const SettingsFileView = Backbone.View.extend({
...
+            if (shouldUpdateFingerprint) {
+                this.model.set('fingerprint', null);
```
antelle: this should be in `file-model::setSyncComplete`

antelle:
Thanks, merged to `develop`.

antelle:
Question: do you want me to cherry-pick it into the release branch, or you build it yourself from develop?

alex-shpak:
HI! Sorry, was on trip.
I think it's not time yet to put it into release,
I will build new version with cordova and will give it a test.

Thank you for merge! 🎉

Issue #223, opened 2016-05-12
Display expired entries when opening or unlocking the database


philogit:
In KeeWeb, the only way to know if an entry is expired is to see it by chance when reviewing a list or search result. I would like KeeWeb to to automatically display expired and soon-to-expire entries as a search result whenever one opens or unlocks the database.

My understanding of the purpose of the field (within KeePass) is to be able to proactively act on an entry the will soon expire, or has expired (e.g., renew an expiring license, expect the arrival of an expiring credit card, etc.). Without an expiry list, the user will never know that an entry has expired unless the user stumbles upon the entry or in the unlikely event that all their entries happen fit within the screen height.

This was also requested as part of #49 (Add security check), however, that would still require the user to remember to periodically perform the check.

===
KeeWeb v1.1.4 (08087c3, 2016-04-21)
Environment: electron v0.37.4
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_11_4) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.1.4 Chrome/49.0.2623.75 Electron/0.37.4 Safari/537.36

antelle:
Audit link will contain a special bubble in case of notifications (like update bell).
The expired entries list will be inside Audit tab. Closing this, as the notification is planned in Audit and we'll not show the expiry list on app startup.

philogit:
That's a sensible plan. Thanks for your response.

PR #693, opened 2017-07-24
Add ability to sort by expired entries

hultberg:
This PR introduces the ability to sort by expired entries via the sort dropdown. I find this helpfull when attempting to find my expired entries. I searched in the issues and noticed that [1] kinda touches in here, but, rather than display expired entries on startup, you can simply sort then later or while searching. In [1] I also noticed there was talk about an Audit log, I'm not sure if this feature is needed, if an audit log is upcoming, but I still see this feature helpful.

1: #223


antelle:
We'll have audit in next version, so it will have an option to search for expired passwords in a slightly different way.

derpferd:
@antelle Has the audit become a thing? If not, I think this is a useful search feature in the mean time.

antelle:
Not yet, but eventually it should

PR #754, opened 2017-10-15
implement Docker support on raspberry pi

tfenster:
I implemented support to run the Docker image on a raspberry pi (armhf-based). I tried to keep it as close to possible to your original Dockerfile and therefore kept installing nginx (as there is no official armhf nginx package) and installing the rest in to separate RUNs. I also had to change the way how daemon off is set in nginx as your sed didn't work for me (there is no pid directive in the standard conf). Lastly I moved the entrypoint copy closer to the end as I played a lot with that script and docker build now only needs to change a few layers in the end if you change that file instead of always downloading from github

### DIFF
--- a/package/docker-rpi/Dockerfile
+++ b/package/docker-rpi/Dockerfile
@@ -1,0 +1,58 @@
+# KeeWeb official docker container
+# https://keeweb.info
+# (C) Antelle 2017, MIT license https://github.com/keeweb/keeweb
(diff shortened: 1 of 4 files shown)


antelle:
🎁  thanks!
About nginx, if there's a built-in webserver in raspberry pi, you can use it, instead of nginx. It's added only to use something as a webserver.
I cannot test it though, because I don't have a raspberry pi.

PR #811, opened 2017-12-17
fix(DetailsHistoryView): small typo

ffflorian:
This PR will fix a small typo in the `DetailsHistoryView`.


antelle:
🎁  thanks!
Wow, these strings should be in locale.js indeed, I'll remove them from here.

PR #843, opened 2018-01-15
Fixes #531: Window icon scaling error on win32 Electron app

### LINKED ISSUE #531: Icon is not quite round
Hi
I'm using the last version of your app and tested in Windows 7, 10 and also Ubuntu and Arch but the result was the same. the app icon in main app and tray system is not not quite round.

kasoban:
Fixes the scaling error reported in #531, caused by some issue with Electron and setting the icon on BrowserWindow on win32.
Instead, if we leave it out like I did now, it will use the icon for the Electron executable currently defined in the Gruntfile.
See https://github.com/atom/atom/issues/4811 for reference.

Current screenshot of result after packaging:
iconscaling

Warning regards testing/validation: This will result in the default Atom icon being displayed when not actually packaging the electron app, e.g. when running 'npm run-script electron'

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -137,16 +137,19 @@ function readAppSettings() {
 
 function createMainWindow() {
     const appSettings = readAppSettings();
-    mainWindow = new electron.BrowserWindow({
+    const windowOptions = {
         show: false,
         width: 1000, height: 700, minWidth: 700, minHeight: 400,
-        icon: path.join(__dirname, 'icon.png'),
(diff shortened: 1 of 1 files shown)


kasoban:
Other electron apps handled this issue by only ever setting the BrowserWindow icon for process.platform linux.
Since I didn't know whether this build makes relevant use of it for darwin, I chose the "not-Windows" switch instead for now.

antelle:
Wow thanks a lot! I didn't manage to find the issue in Atom you mentioned, and I thought something was wrong with the icon itself. 
🎁

PR #866, opened 2018-02-08
Fix #456: Add more logs around PopupNotifier & handle crash & did-fai…

### LINKED ISSUE #456: Sync to Google Drive spinning forever
### Actions:
1. the sync process starts automatically (not started by user)
2. the sync process is stuck forever, this only happens sometimes, I'm not sure what causes it

### Investigation:

By looking the logs I've noticed that successful syncs to Google Drive usually have a "OAuth popup opened" and are followed by a "OAuth success". When the sync gets stuck the latter doesn't happen. For example:
```
2016-11-30T18:40:28.856Z [storage-gdrive] OAuth popup opened
2016-11-30T20:10:28.093Z [updater] Checking for update...
```

So looking at the code I see that one thing that can fail is opening the popup so I opened devtools and tried calling `window.open` with some parameters, which caused a window to open, when I closed it then the sync process got unstuck:

```
var url = 'https://accounts.google.com/o/oauth2/v2/auth';
var title = 'OAuth';
var settings = "width=600,height=400,left=1166,top=116,dialog=yes,dependent=yes,scrollbars=yes,location=yes";
var foo =  window.open(url, title, settings);
2016-12-02T02:29:57.853Z [storage-gdrive] OAuth error popup closed
2016-12-02T02:29:57.853Z [storage-gdrive] Stat error 47264085ms unauthorized
2016-12-02T02:29:57.854Z [sync:Passwords v2] Stat error, dirty, save to cache unauthorized
```

Then I was able to click in "Save" and the sync happened successfully.

### Expected behavior: 
Sync start automatically and finish within some time limit (e.g. 30 secs)

### App information:
```
KeeWeb v1.3.3 (52701af, 2016-09-20)
Environment: electron v1.3.3
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_12_1) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.3.1 Chrome/52.0.2743.82 Electron/1.3.3 Safari/537.36
```

Let me know if you need more information. Thanks!

tanob:
…l-load events

🎁

### DIFF
--- a/app/scripts/comp/popup-notifier.js
+++ b/app/scripts/comp/popup-notifier.js
@@ -5,11 +5,16 @@ const AuthReceiver = require('./auth-receiver');
 const Links = require('../const/links');
 const Timeouts = require('../const/timeouts');
 const Locale = require('../util/locale');
+const Logger = require('../util/logger');
 
 const PopupNotifier = {
+    logger: null,
+
     init: function() {
+        this.logger = new Logger('PopupNotifier');
+
         if (Launcher) {
-            window.open = this._openLauncherWindow;
+            window.open = this._openLauncherWindow.bind(this);
         } else {
             const windowOpen = window.open;
             window.open = function() {
@@ -66,7 +71,16 @@ const PopupNotifier = {
                 PopupNotifier.processReturnToApp(toUrl);
             }
         });
-        win.loadURL(url);
+        win.webContents.on('crashed', (e, killed) => {
+            this.logger.debug('crashed', e, killed);
+            setTimeout(PopupNotifier.triggerClosed.bind(PopupNotifier, win), Timeouts.CheckWindowClosed);
+            win = null;
+        });
+        win.webContents.on('did-fail-load', (e, errorCode, errorDescription, validatedUrl, isMainFrame) => {
+            this.logger.debug('did-fail-load', e, errorCode, errorDescription, validatedUrl, isMainFrame);
+            setTimeout(PopupNotifier.triggerClosed.bind(PopupNotifier, win), Timeouts.CheckWindowClosed);
+            win = null;
+        });
         win.once('page-title-updated', () => {
             setTimeout(() => {
                 if (win) {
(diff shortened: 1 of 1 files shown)


app/scripts/comp/popup-notifier.js
```diff
@@ -66,7 +71,16 @@ const PopupNotifier = {
...
         });
-        win.loadURL(url);
+        win.webContents.on('crashed', (e, killed) => {
+            this.logger.debug('crashed', e, killed);
+            setTimeout(PopupNotifier.triggerClosed.bind(PopupNotifier, win), Timeouts.CheckWindowClosed);
+            win = null;
+        });
+        win.webContents.on('did-fail-load', (e, errorCode, errorDescription, validatedUrl, isMainFrame) => {
+            this.logger.debug('did-fail-load', e, errorCode, errorDescription, validatedUrl, isMainFrame);
+            setTimeout(PopupNotifier.triggerClosed.bind(PopupNotifier, win), Timeouts.CheckWindowClosed);
```
antelle: I'm a bit concerned about this. We're triggering `closed` event here, however the window is not closed. Won't it result in multiple windows being opened because the app thinks there's no active popup?

app/scripts/comp/popup-notifier.js
```diff
@@ -5,11 +5,16 @@ const AuthReceiver = require('./auth-receiver');
 const Links = require('../const/links');
 const Timeouts = require('../const/timeouts');
 const Locale = require('../util/locale');
+const Logger = require('../util/logger');
 
 const PopupNotifier = {
+    logger: null,
+
     init: function() {
+        this.logger = new Logger('PopupNotifier');
```
antelle: please replace with `new Logger('popup-notifier')`

app/scripts/comp/popup-notifier.js
tanob: Electron doesn't actually open a window for the cases I've detailed in #456 so I didn't experience any issues with this change. I see you changed to call `win.close`, I think that's good so at least it can free any resources being held.

tanob:
Thanks for merging & doing the adjustments, I got busy with other things and couldn't do them :-)

antelle:
🎁  thanks for the pr! I'll add more storage and popups fixed in the next version, so it will be helpful.

PR #871, opened 2018-02-14
Add support to pt-BR

amsbr:
Add native support to Portuguese Brazilian language


antelle:
I'm sorry, but why is that? We support DE and FR, other language are in plugins.

amsbr:
Sorry, I just found pt-pt. I have now seen that it has pt-br.

Thanks

antelle:
Yes :) We have two languages.
Oh, you translated the whole thing to pt-br? I thought you just moved it here from plugins.keeweb.info for whatever reason.
That's huge amount of work! Nut yes, we already have it translated...  You can improve the existing translations on OneSky: http://keeweb.oneskyapp.com

Issue #978, opened 2018-08-15
KDE: KeeWeb shown among active windows even if minimized to tray


ceres-c:
Hello, I believe I have found a problem with KDE and KeeWeb. See attached images for explanation.
1) KeeWeb is first opened, as you can see everything is normal and there is no icon in the right tray
![1](https://user-images.githubusercontent.com/24912818/44145285-4842c5d0-a08b-11e8-9856-7fcc6b8eb43f.png)
2) KeeWeb is closed to tray (still OK)
![2](https://user-images.githubusercontent.com/24912818/44145322-6d1690c6-a08b-11e8-9017-3cafe8c5a18b.png)
3) KeeWeb is again brought on top by alt-tabbing and cycling through every window
![3](https://user-images.githubusercontent.com/24912818/44145420-c7d8f18e-a08b-11e8-8daa-86cd0fe00b42.png)
4) KeeWeb is once again minimized to tray and another keeweb icon appears in the tray, effectively duplicating the original one. There is no way to delete this second icon now.
![4](https://user-images.githubusercontent.com/24912818/44145460-f726ab48-a08b-11e8-9391-ecd4f0b39a79.png)

This can be fixed by setting KDE to not cycle through minimized windows.
![5](https://user-images.githubusercontent.com/24912818/44145535-3b219812-a08c-11e8-80d4-1953c136f1f4.png)

Still, I do not believe this is the correct behavior. I have never developed with electron, so I can't propose a fix, sorry.

antelle:
Hi! Thanks for the detailed explanation. So, it looks like hiding a window still leaves it available, while we think it's not visible. In this state we assume that the only way to show it is through interacting with the icon, that's why the second icon is inserted.

Coises:
I can confirm that the same thing happens on Windows 7.

In Windows, the typical thing to do when “minimizing” to the tray would be to hide the window rather than to minimize and suppress the taskbar entry (lines 92-93 in release-1.6/desktop/app.js — restoreMainWindow would also have to change to handle hidden windows and not just minimized ones).

That keeps the application from appearing in the alt+Tab list while it’s “minimized” to the tray, which is expected behavior on Windows. I don’t know if Linux users expect the same thing, or if Linux behaves the same way.

ceres-c:
Yes, I expect it not to be shown in the alt-tab list while minimized, to answer to your question. This is the correct behavior.

[...]

PR #979, opened 2018-08-15
Proper minimize to tray for Windows (and Linux?)

Coises:
On Windows 7, this corrects the issue described in Issue #978.

Commit https://github.com/keeweb/keeweb/commit/d76cc891e17c93efc2b52de37db69bb28fd37792 broke minimize to tray on Windows: the tray icon appeared, but the window didn’t go away. This fixes that problem as well. The error appeared to be a result of a simple omission in the `minimizeApp` function when the Darwin and non-Darwin code paths were split; `MainWindow.minimize();` was left out of the non-Darwin path. The fix for both issues is to replace `MainWindow.minimize();` and `mainWindow.setSkipTaskbar(true);` with `mainWindow.hide();` in the non-Darwin path, and to replace `mainWindow.focus();` with `mainWindow.show();` in the `restoreMainWindow` function. This produces expected Windows behavior.

I have no way of testing this except on Windows 7. I had to base on develop branch since I see no way to build master on Windows.

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -93,10 +93,11 @@ app.minimizeApp = function () {
         mainWindow.hide();
         app.dock.hide();
         imagePath = 'mac-menubar-icon.png';
+        mainWindow.setSkipTaskbar(true);
     } else {
+        mainWindow.hide();
         imagePath = 'icon.png';
     }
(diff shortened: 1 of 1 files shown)


antelle:
Thanks! 🎁

Issue #970, opened 2018-08-02
[Feature request] Drag field name, drop field value


Coises:
In both the desktop application and the web app, it would be good if dragging a field name in the details display for an item and dropping it on an input box in another page or application would drop the field value.

Presently, clicking a field name copies the field value to the clipboard, but attempting to drag a field name just highlights surrounding text, which isn’t useful. It’s possible now to triple-click in the field value input area, pause, then drag the highlighted text to get the desired effect; it would just be smoother and more convenient simply to drag the field by the field name.

The reason drag and drop is useful is that sometimes you might need to use multiple fields from an item in a context that’s too unpredictable for auto type. An example would be filling in credit card information: the cardholder’s name, billing address, credit card number and CVV2 will all be needed, but they won’t be in any predictable configuration on the page. To copy and paste, one must repeatedly switch back and forth between KeeWeb and the target web page; drag and drop is less clumsy. Drag and drop also avoids putting sensitive information on the clipboard.

Hopefully this would be simple and straightforward using HTML5 drag-and-drop.

matheod:
https://developer.mozilla.org/en-US/docs/Web/API/HTML_Drag_and_Drop_API/Drag_operations

PR #985, opened 2018-08-23
Draggable field labels

Coises:
This implements the change requested in #970 by allowing field labels in the details view to be dragged. The value dragged is the same as the value copied to the clipboard by clicking on the label.

In testing I saw that as soon as the label began dragging, the “drop attachments here” message replaced the details view. While this didn’t interfere with dragging the field, it was somewhat jarring and unexpected.

Further investigation showed that in many places KeeWeb indicates that it is possible to drop something without first checking what is being dragged. https://github.com/keeweb/keeweb/commit/7a3cc9c863ce4a8460e83d925225638a97b53ea4 limits the “drop attachments here” message to appear only when a *file* is dragged across the details view. https://github.com/keeweb/keeweb/commit/2450e3a0b607203b9c7687424cff7dbd0d57f6d5 makes KeeWeb indicate that something can be dropped only when dragging an item of a type which can be dropped there. https://github.com/keeweb/keeweb/commit/0f1a248fcc7347aa52c6953c75bc7d58925123bf implements draggable field labels.

I’ve tested this only on Windows 7 with the desktop client and with Firefox, Chrome and Vivaldi.

### DIFF
--- a/app/scripts/views/app-view.js
+++ b/app/scripts/views/app-view.js
@@ -31,6 +31,7 @@ const AppView = Backbone.View.extend({
     events: {
         'contextmenu': 'contextMenu',
         'drop': 'drop',
+        'dragenter': 'dragover',
         'dragover': 'dragover',
         'click a[target=_blank]': 'extLinkClick',
         'mousedown': 'bodyClick'
@@ -678,6 +679,7 @@ const AppView = Backbone.View.extend({
 
     dragover: function(e) {
         e.preventDefault();
+        e.originalEvent.dataTransfer.dropEffect = 'none';
     },
 
     drop: function(e) {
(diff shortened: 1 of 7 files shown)


antelle:
Thanks a lot! 🎁 
I tested it on Mac, and I can drag fields outside of the app, it creates a file in this case. They can't be dragged on text fields in other apps though, but maybe that's how it works.
Anyway, it doesn't break anything and seems to be valid.

Coises:
>I tested it on Mac, and I can drag fields outside of the app, it creates a file in this case. They can't be dragged on text fields in other apps though, but maybe that's how it works.

I’m not familiar with Mac, but this doesn’t seem like it should be. If you highlight some text in an ordinary web page (like this one), can you drag into a text field on a Mac? If that works, we should be able to make this work.

Just a guess, but possibly in app/scripts/views/fields/field-view.js line 104, Mac wants "text" instead of "text/plain"?

If that makes it work on Mac, I can check to see if it still works on Windows. It was unclear to me whether "text" or "text/plain" is the more proper universal descriptor.

Thanks for looking at this, Antelle.

Issue #646, opened 2017-05-30
Master password should be inserted twice to check correctness


andreapx:
- what was wrong? I think that for changing the "Master password" in the settings, it should be inserted twice to avoid errors writing it. Yesterday for example I must accidentally have added a char to that field just before saving the file and...the file is now not anymore accessible.
- app version: **1.5.2**
- your user-agent (from Settings/Help section): Mozilla/5.0 (X11; Ubuntu; Linux x86_64; rv:53.0) Gecko/20100101 Firefox/53.0

Thanks!

antelle:
you can create a plugin for it

andreapx:
I would like to be able to do it!!!

gynet:
I think double checking the input during changing the master password should be important especially for someone like me has a fat finger :)

antelle:
Well, master password is a bit special compared to other passwords, let's add this to the core product.

andreapx:
Can I have hire a freelance to solve this and then upload it here?

antelle:
It's up to you, if you want it right now, of course you can, but please note that there may be some changes requested. As I see it now: when you change your password, we show an additional text field below it, where you can repeat your password.

andreapx:
What do you mean by "there may be some changes requested"?

antelle:
Whan you submit a pull request on github, project maintainersay want to request changes to be made, i.e. if there's something not implemented properly (not alighed with maintainers' point of view) in the pr: design, architecture, etc...

andreapx:
Is there anybody willing to do this "now" upon payment?

austil:
I looked at this last friday, the changes required seems fairly simple.
I may have the time to submit a pull request before the end of the year.

And last but not least, I'm willing to do this for free. This feature is important to me (never changed my master password because of this). Feel free to make a donation to the maintainer of this pretty useful app instead 😉

andreapx:
That's great!!!
Let me know when you upload it!

andreapx:
@austil 
When you'll do it, will it go directly in app.keeweb.info?

antelle:
@andreapx after a successful merge it will be included in the next release and then yes, to app.keeweb.info and apps, after some time, together with other features. So, not directly but eventually it will be there.

austil:
As you probably noticed, I haven't found time to do it. I'm not in control of my time at the moment and will not be before June ...
[...]

PR #988, opened 2018-08-27
#646 Implement confirm master password

sahalsaad:

Please follow these steps:

1. work on the right branch: it must be `develop` for new features and `master` for hotfixes
2. make sure the feature you add is not listed in [banned features](https://github.com/keeweb/keeweb/wiki/Unsupported-Features)
3. check compatibility with [the project roadmap](https://github.com/keeweb/keeweb/wiki/TODO) and key features: you should not break anything
4. if your feature is introducing a lot of new UI, or is changing any UX or design concept, please, open an issue to discuss it first
5. please, respect existing style in code, styles and markup (hint: install and enable [editorconfig](http://editorconfig.org/) in your editor)
6. don't add any dependencies
7. test your code: it must work in browser, Windows, OSX and Linux
8. if your code is platform-dependent, please add corresponding switches
9. respect each platform: don't create behaviour which breaks interface guidelines or user expectations 
10. run `grunt` and make sure it's passing 
11. Implement confirm master password #646
screenshot from 2018-08-27 09-48-06
screenshot from 2018-08-27 09-48-22


app/scripts/locales/base.json
```diff
@@ -428,6 +429,8 @@
   "setFileDontUseKeyFile": "Don't use key file",
   "setFileEmptyPass": "Empty password",
   "setFileEmptyPassBody": "Saving database with an empty password makes it completely unprotected. Do you really want to do it?",
+  "setFilePasswordError": "Error updating password",
+  "setFilePasswordErrorBody": "Master password does not match the confirm master password.",
```
antelle: "confirm master password" should be probably something else, like "confirmed password", or something that make a sentence

app/scripts/models/file-model.js
```diff
@@ -15,6 +15,7 @@ const FileModel = Backbone.Model.extend({
         name: '',
         keyFileName: '',
         passwordLength: 0,
+        confirmPasswordLength: 0,
```
antelle: Please don't change the model for this feature, it must live in the UI and change the model when both passwords match.

app/scripts/views/settings/settings-file-view.js
```diff
@@ -9,7 +9,6 @@ const Links = require('../../const/links');
 const Format = require('../../util/format');
 const Locale = require('../../util/locale');
 const UrlUtil = require('../../util/url-util');
-const kdbxweb = require('kdbxweb');
```
antelle: Same as above, I guess, it changes almost everything, please open a new PR then.

app/scripts/views/settings/settings-file-view.js
```diff
@@ -143,6 +146,17 @@ const SettingsFileView = Backbone.View.extend({
             });
             return false;
         }
+
+        if (this.model.get('passwordChanged') && !this.model.get('passwordMatched')) {
+            Alerts.error({
```
antelle: This should be displayed inline, next to the confirmation field

app/scripts/views/settings/settings-file-view.js
```diff
@@ -382,12 +396,27 @@ const SettingsFileView = Backbone.View.extend({
     blurMasterPass: function(e) {
         if (!e.target.value) {
             this.model.resetPassword();
-            e.target.value = PasswordGenerator.present(this.model.get('passwordLength'));
+            const password = PasswordGenerator.present(this.model.get('passwordLength'));
+            e.target.value = password;
             this.$el.find('.settings__file-master-pass-warning').hide();
+            this.$el.find('#settings__file-confirm-master-pass').val(password);
```
antelle: Why? The confirmation field must be empty when the focus leaves the master password field, with a message that confirmation is required.

app/scripts/views/settings/settings-file-view.js
sahalsaad: What happen if user click save button if the password entered not match?

app/scripts/views/settings/settings-file-view.js
sahalsaad: Because of this `leave the field blank to use the old password`.

Issue #877, opened 2018-02-27
Issue with WebDav


LiloBzH:
- what were your actions? Sync with Webdav and Owncloud
- what was wrong? Erreur 500 . Keeweb send a request with url : file://
- app version 1.6.3 (windows)
- does it happen on Demo or New database? all
- please, open dev tools in your browser and attach output log from Console tab 
    (if you are using desktop app, devtools can be opened from Settings/General/Advanced)

```
[DEBUG] 2018-02-27T08:40:33.178Z [storage-webdav] Save:put error https://cloud.blop.com/remote.php/dav/files/blop/.blop.kdbx.4851678186768 500 337ms
[INFO ] 2018-02-27T08:40:33.178Z [sync:Lilo] Error saving data to storage
[INFO ] 2018-02-27T08:40:33.179Z [sync:Lilo] Sync finished HTTP status 500
```
Ticket Owncloud here : https://github.com/owncloud/core/issues/30330 but we think the problem is Keeweb.

Keeweb seems bo be send local URL first and after remote url

first : file://
2nd : https://cloud.blop.com/remote.php/dav/files/blop/.coucou.1519646962149
3rd : file://
4 : https://cloud.blop.com/remote.php/dav/files/blop/.coucou.1519646962149

antelle:
Hi! This is a problem with your webserver, somehow it returns status 500. You can go to server logs to check what's happening.

LiloBzH:
are you serious ?

If i open a ticket, it's not a problem of webserver. Can you read my ticket ?

antelle:
So, what's the problem in keeweb? You get 500 from your server. Did you figure out, why? We send an incorrect url, or what happens?

LiloBzH:
Webdav Synology <-> OWncloud is OK. No problem.

log of owncloud

[...]

PR #990, opened 2018-08-28
Fix for Issue #877 - remove bad Origin header

Coises:
I believe this will fix [#877 Issue with WebDav](https://github.com/keeweb/keeweb/issues/877).

It does need more testing than I can give it. It appears to make my OwnCloud installation work with KeeWeb, and it doesn’t appear to interfere with access to Google Drive. The changes are all in desktop/app.js, so the web version isn’t affected at all.

The underlying problem is that Chromium sends a mal-formed Origin header when it sees a PUT XMLHttpRequest from a local file. (Chromium sends `Origin: file://`; per [rfc6454](https://tools.ietf.org/html/rfc6454#section-7), only scheme://host and scheme://host:port are valid; if the origin does not have a scheme and a host, the literal value `none` must be used if the Origin header is supplied.)

The Origin header is supplied by the browser and can’t be rewritten using XMLHttpRequest; fortunately, Electron provides a global mechanism for trapping outgoing network requests via [webRequest](https://electronjs.org/docs/api/web-request).

### DIFF
--- a/desktop/app.js
+++ b/desktop/app.js
@@ -60,6 +60,7 @@ app.on('ready', () => {
         setGlobalShortcuts();
         subscribePowerEvents();
         deleteOldTempFiles();
+        hookRequestHeaders();
     }
 });
 app.on('open-file', (e, path) => {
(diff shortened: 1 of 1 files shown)


antelle:
Yeah this will fix it in desktop, thanks!

acabutto:
Hi, any chance to have this fix merged into master? Or a publicly available develop build?
I unsucessfully tried to build my own develop release.

antelle:
@acabutto dev build is available here: https://beta.keeweb.info

Coises:
@antelle: Is there some way to download a desktop build from beta.keeweb.info? If so, I don’t see it (so I’m guessing others might not, either).

This is a desktop-only fix. In a browser, the origin header is appropriate (and we couldn’t change it even if we wanted to). If KeeWeb isn’t hosted on the same domain as the WebDAV server, the WebDAV server has to accept the relevant CORS requests.

antelle:
Currently there's no way to download beta desktop app.

Issue #807, opened 2017-12-13
[Feature request] Option to force single instance mode


bnbaz:
I often find myself with KeeWeb open (Windows 10), then hitting the windows-key, typeing 'keeweb', hitting enter, and logging in the second time. Of course, I just wanted to use the KeeWeb instance that was already running.

Therefore, it would be very nice if there was an option to force a single instance, so the running instance is brought to the front when the user attempts to start KeeWeb while it is already running.

antelle:
Hi!
KeeWeb is running single-instance mode, maybe there's a bug?
Maybe you have "Lock on minimize" option selected, it just gets minimized, and then you see the same instance? Or it really starts a new process?

bnbaz:
Looks like a bug then. The app is not minimized, it appears regardless of
whether or not I have 'lock on minimize' selected.

The log doesn't show anything funny to me. I shortened the identifiers but
didn't remove anything else from the log. This is happening on a Windows 10
64-bits system, fully up to date (Fall Creators).

[INFO ] 2017-12-14T08:28:19.913Z [app] Started in 915ms ¯\_(ツ)_/¯
[INFO ] 2017-12-14T08:28:19.916Z [updater] Next update check will happen in
84489s
[INFO ] 2017-12-14T08:28:23.894Z [open:keepa] File open request
[INFO ] 2017-12-14T08:28:23.894Z [open:keepa] Open file from cache, will
sync after load gdrive
[DEBUG] 2017-12-14T08:28:23.895Z [storage-cache] Load 402993bf-
[DEBUG] 2017-12-14T08:28:23.897Z [storage-cache] Loaded 402993bf- 1ms
[INFO ] 2017-12-14T08:28:23.898Z [open:keepa] Loaded file from cache null
[INFO ] 2017-12-14T08:28:24.225Z [file] Opened file ke: 325ms, 6000 rounds,
21 kB
[DEBUG] 2017-12-14T08:28:24.225Z [app] Add last open file 402993bf- ke
gdrive 0BzF- 0BzF-
[INFO ] 2017-12-14T08:28:24.320Z [open:keepa] Sync just opened file
[INFO ] 2017-12-14T08:28:24.332Z [sync:keepa] Sync started gdrive 0BzF- {}
[INFO ] 2017-12-14T08:28:24.334Z [sync:keepa] Stat file
[DEBUG] 2017-12-14T08:28:24.336Z [storage-gdrive] Stat 0BzF-
[...]

PR #993, opened 2018-09-02
Fix Issue #807 - single instance not enforced

Coises:
This fixes issue #807 on Windows 7 and presumably elsewhere. All that was wrong is that we were checking for a previous instance in the app.on('ready') handler. We need to check in code that runs before that event gets raised.


antelle:
Thanks a lot! :gift:

PR #994, opened 2018-09-03
Adjust some overflow behavior in the details panel, includes fixing Issue #945

### LINKED ISSUE #945: Navigation in version history impeded by long comment
- Put a REALLY long (> 200 chars) text in the "Notes" section (or is it "Comment" in the english version?) of an entry.
- Create several versions of this entry, some with a time stamp very long ago (> 1 year)
- go to the version history and navigate back (left) to the oldest version
==> the arrow right and buttons "restore this version" "delete this version" have moved off the panel to the right. Hence no navigation back to a more recent version is possible (except stepping out and back in to this entry) 
The problem would be avaioded if the Comment was also displayed with lines wrapped like in the normal display.
Long text in the comment field IMO should not be considered a corner case.
-------------
KeeWeb v1.6.3 (cded8a4, 2017-12-11)
Environment: electron v1.7.9
User-Agent: Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.6.3 Chrome/58.0.3029.110 Electron/1.7.9 Safari/537.36

Coises:
Fix various glitches in details panel layout:

* Make long Website fields have the same width as other fields, and leave more than a sliver of space at the end in which to click to edit.

* Let the attachment download/delete instructions wrap if they are too long to fit on one line.

* Make fields in the history wrap the same way they do in the ordinary details pane (fixes #945).

* Avoid having the navigation area and the top of the history details overlap when the window height is too small to show everything without scrolling.

* Align the right edges of the history navigation area and the revert/delete buttons so that they don't overlap the scroll bar.

### DIFF
--- a/app/styles/areas/_details.scss
+++ b/app/styles/areas/_details.scss
@@ -264,6 +264,12 @@
         font-weight: normal;
         @include user-select(none);
       }
+      >a {
+        max-width: 100%;
+        white-space: nowrap;
+        overflow: hidden;
+        text-overflow: ellipsis;
+      }
       .details__body-aside & {
         @include th { color: th(muted-color); }
         a { @include th { color: th(muted-color); } }
@@ -402,7 +408,6 @@
       position: absolute;
       bottom: $base-padding-v;
       right: $base-padding-h;
-      white-space: nowrap;
       opacity: .15;
       display: none;
       @include nomobile { display: block; }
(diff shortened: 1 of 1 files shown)


antelle:
Thanks! 🎁

PR #995, opened 2018-09-05
tray-min-auto-type-select-fix

Coises:
The fix for alt-tab behavior when KeeWeb is minimized to the tray in 3dae878 left a problem when auto-type raises a selection list: the taskbar button shows, and after a selection is made KeeWeb minimizes
to the taskbar but leaves a tray icon present. The same thing happens if auto-type is canceled by clicking either the minimize button or the close button at the top right of the selection window. From this state, various scenarios lead to having duplicate tray icons.

This fix restores the behavior of 1.6.3 when auto-type raises a selection list while KeeWeb is minimized to the tray: the selection window shows, the tray icon stays, and no taskbar button shows.

There remains a case which is not handled perfectly: when KeeWeb is minimized to the tray, auto-type raises a selection list, and the user cancels auto-type by some means other than clicking the close button. This is a bit of a "corner case," the resultant behavior is "odd" but not unusable, and it would have to be handled in a different code path from the one affected by this commit, so I didn't attempt to include it in this fix.

Further details are in the commit comment.

### DIFF
--- a/app/scripts/comp/launcher-electron.js
+++ b/app/scripts/comp/launcher-electron.js
@@ -200,7 +200,7 @@ const Launcher = {
     hideApp: function() {
         const app = this.remoteApp();
         if (this.canMinimize()) {
-            app.getMainWindow().minimize();
+            app.minimizeThenHideIfInTray();
         } else {
             app.hide();
         }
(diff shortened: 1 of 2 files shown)


antelle:
Thanks! 🎁  works well on mac as well.

Issue #954, opened 2018-07-05
Keeweb Desktop (Windows) stuck on unplugged monitor


tbradsha:
I traditionally have a two-monitor setup. After putting the laptop to sleep and unplugging the monitor, the Keeweb window is stuck on the non-existent monitor. Killing the app and restarting it doesn't help; it keeps respawning on the other monitor.

Until this bug is squashed, is there a workaround to somehow reset the preferences (or better yet, edit some config file)?

mitcom:
I don't know how KeeWeeb is rendered and displayed, but have you tried to move that window from other monitor to your main one? I mean by using [Win] + [Arrow Left] few times

tbradsha:
**Workaround:** [AltKey] + [Space] will activate the shortcut menu, and then selecting "Move" allows you to move the window to the new screen.

Yes, I tried using [WinKey] + [Left Arrow], but to no avail. Also odd is that the [WinKey] + [Tab] doesn't pull up the "Desktop" view (not sure the terminology in Windows 10) unless the app is either closed or in the notification tray (not just minimized).

[AltKey] + [Tab] shows the window thumbnail, so I can at least see if it's prompting me to log in or if there's an error, but I can't drag it around in that mode.

Coises:
Probable workaround (I don’t have a second monitor handy, so I can’t test):

* Close KeeWeb by any means necessary.
* Type **WinKey**+**R**, enter `%appdata%\KeeWeb` and click OK.
* Delete the `window-position.json` file.
* Restart KeeWeb.

I see that GitHub has [a similar problem](https://github.com/desktop/desktop/issues/2107) which was reported over a year ago and still hasn’t been fixed. I haven’t looked at the relevant Electron APIs yet... it’s possible that Electron doesn’t expose the information we need to solve this, or maybe the GitHub folks just haven’t had time to fix it.

Coises:
@tbradsha: Does it make any difference if you actually disable the second monitor (**WinKey**+**P** and select main screen only) before you disconnect it?

tbradsha:
@Coises Removing the window-position.json file does the trick, so that's a great workaround.

[...]

PR #1002, opened 2018-09-23
always-restore-to-a-connected-monitor (addresses Issue #954)

Coises:
Issue #954: If a display is disconnected while KeeWeb is minimized,
when the window position is restored (either during the same session
or after quitting and restarting KeeWeb) it will typically be positioned
where the now disconnected display used to be, making it difficult
for the user to find the window and reposition it to a connected display.

This commit makes KeeWeb check whenever the window position is restored
to be sure there is enough of the title bar on some connected display
that the user can find and drag the window. If there is not, the window
is repositioned to the primary display.

I've been able to test this only on Windows 7, using one and two monitors.


medozas:
Test with this PR was successful for me.

PR #1013, opened 2018-10-08
Fix the link of list of issues which need help

ko-dever:
This PR fixes the links of issues which need help, as the old label `need help` seems to no longer exist.


antelle:
Hi! It's fixed in `develop`

PR #1018, opened 2018-10-12
Implemented hotkeys (option/enter and command/enter) to support separ…

dploeger:
This PR isn't really meant to be merged as such.

I have some things to clear first, but I'm not quite accustomed to the code and need your help for this:

* I think, adding entry.autoTypeOption is a bit to pragmatic, right? I guess, I should add it to some model first before using it.
* This feature isn't documented anywhere. I would propose to add a small hint in the auto type selector and/or document it in the shortcuts view. What do you think?

btw: thanks for this great app!

### DIFF
--- a/app/scripts/comp/key-handler.js
+++ b/app/scripts/comp/key-handler.js
@@ -12,7 +12,7 @@ const KeyHandler = {
     modal: false,
 
     init: function() {
-        $(document).bind('keypress', this.keypress.bind(this));
+        $(document).bind('keyxpress', this.keypress.bind(this));
         $(document).bind('keydown', this.keydown.bind(this));
 
         this.shortcuts[Keys.DOM_VK_A] = [{ handler: this.handleAKey, thisArg: this, shortcut: this.SHORTCUT_ACTION,
--- a/app/scripts/comp/key-handler.js
+++ b/app/scripts/comp/key-handler.js
@@ -12,7 +12,7 @@ const KeyHandler = {
     modal: false,
 
     init: function() {
-        $(document).bind('keyxpress', this.keypress.bind(this));
+        $(document).bind('keypress', this.keypress.bind(this));
         $(document).bind('keydown', this.keydown.bind(this));
 
         this.shortcuts[Keys.DOM_VK_A] = [{ handler: this.handleAKey, thisArg: this, shortcut: this.SHORTCUT_ACTION,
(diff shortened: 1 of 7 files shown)


antelle:
Hi! Yes, `entry.autoTypeOption` won't work indeed. I would suggest adding a flag that will be passed together with current entry [here](https://github.com/keeweb/keeweb/blob/master/app/scripts/auto-type/index.js#L197). So, `result` will be `{ entry: ..., sequence: ... }`. The result is [here](https://github.com/keeweb/keeweb/blob/master/app/scripts/views/auto-type/auto-type-select-view.js#L20) and it's initialized on different key press events. You'll need to put `entry` and `sequence` there.

dploeger:
Thanks for checking it out. I'm not quite sure, if I followed, but I moved some code around. Hope, that I understood you correctly.

app/scripts/comp/key-handler.js
```diff
@@ -12,7 +12,7 @@ const KeyHandler = {
     modal: false,
 
     init: function() {
-        $(document).bind('keypress', this.keypress.bind(this));
+        $(document).bind('keyxpress', this.keypress.bind(this));
```
antelle: what is it?

app/scripts/views/auto-type/auto-type-select-view.js
```diff
@@ -26,6 +27,8 @@ const AutoTypePopupView = Backbone.View.extend({
         this.listenTo(Backbone, 'main-window-will-close', this.mainWindowWillClose);
         KeyHandler.onKey(Keys.DOM_VK_ESCAPE, this.escPressed, this, false, true);
         KeyHandler.onKey(Keys.DOM_VK_RETURN, this.enterPressed, this, false, true);
+        KeyHandler.onKey(Keys.DOM_VK_RETURN, this.actionEnterPressed, this, KeyHandler.SHORTCUT_ACTION, true);
+        KeyHandler.onKey(Keys.DOM_VK_RETURN, this.optEnterPressed, this, KeyHandler.SHORTCUT_OPT, true);
```
antelle: Please add an explanation text in the view somewhere. Nobody is going to notice it otherwise.

antelle:
Thanks! 🎁  💯 
Merged it

dploeger:
Awesome! Thanks!

Issue #450, opened 2016-11-22
Please provide a snap package


0xBADEAFFE:
It would be great if you could provide a snap package for keeweb. Instructions are here: http://snapcraft.io/.

owais:
This would be really great. A deb or ppa package would be nice as well but creating a snap is much simpler and it can work on all distros.

piclemx:
It's will give you more visibility with this feature.

pedzed:
Any update on this?

Bodup:
This would be great since snaps also sandbox the application. Snaps finally allow for a somewhat trusted Linux desktop system. (assuming Wayland is used as the display server, rather than X11)

gustawdaniel:
I found tutorial how tho add snap

> https://docs.snapcraft.io/electron-apps/6748

I go through process of running this project on my local machine #1035 and #1038

Then after command  `npm run dist` had error

```
  ⨯ snapcraft is not installed, please: sudo snap install snapcraft --classic
```

Fexed by 

```
sudo snap install snapcraft --classic
```

that installed version

```
snapcraft 2.43.1 from Canonical✓ installed
```

And again had error

```
> keeweb@1.6.3 dist /home/daniel/exp/keeweb
> build --linux snap

  • electron-builder version=20.36.2
  • writing effective config file=dist/builder-effective-config.yaml
  • rebuilding native production dependencies platform=linux arch=x64
  • packaging       platform=linux arch=x64 electron=1.8.8 appOutDir=dist/linux-unpacked
  • building        target=snap arch=x64 file=dist/keeweb_1.6.3_amd64.snap
  • default Electron icon is used reason=application icon is not set
  • application Linux category is set to default "Utility" reason=linux.category is not set and cannot map from macOS docs=https://electron.build/configuration/configuration#LinuxBuildOptions-category
  ⨯ at least snapcraft 2.40.0 is required, please: sudo snap install snapcraft --classic
Error: /home/daniel/exp/keeweb/node_modules/app-builder-bin/linux/x64/app-builder exited with code 1
    at ChildProcess.childProcess.once.code (/home/daniel/exp/keeweb/node_modules/builder-util/src/util.ts:244:14)
    at Object.onceWrapper (events.js:317:30)
    at emitTwo (events.js:126:13)
    at ChildProcess.emit (events.js:214:7)
    at maybeClose (internal/child_process.js:925:16)
    at Process.ChildProcess._handle.onexit (internal/child_process.js:209:5)
[...]

PR #1037, opened 2018-11-18
Npm install fixed #1035 - added dependency to README.md

gustawdaniel:
I added dependency that allow to execute `npm install` without errors.

PR #1039, opened 2018-11-18
 Snap added. #450

gustawdaniel:
I added packages and script in package.json that create snap.

+ Where is documentation for developers?

I suggest replace my icon because of of I used convert and lost quality. 

+ Where is svg icon?
+ Where are tests?
+ What do you think about `npm update`?
+ What do you think about continuous integration?

Who should create account on snapcraft? You or I? If I then I would like to add travis to this project and automate deploying.


gustawdaniel:
graphics/icon.ico it other than desktop/icon.png

antelle:
Hi!

Please include these changes into #1037, closing this, I'll add some comments regarding implementation there.

Answers to questions:

> Where is documentation for developers?

There's something here, but not much: https://github.com/keeweb/keeweb/wiki
> Where is svg icon?

It doesn't exist
> Where are tests?

We don't have tests
> What do you think about npm update?

Will do this right before next release.
> What do you think about continuous integration?

Building apps is impossible because of code signing. CI by itself doesn't make sense if we don't have tests.

PR #1050, opened 2018-12-18
Fixing issue where URL loaded plugins fail to automatically load after program restart

thecount2a:
Fixing issue where plugin manager does not persist any info across program restarts about whether a particular plugin was a user-installed plugin (and therefore is exempt from signature checking). This will allow URL installed plugins to stay across restarts of the program.


thecount2a:
Just realized I applied this to the wrong branch

Issue #1042, opened 2018-11-19
Custom theme not loaded on refresh


crivotz:
I have configured a local version of keeweb on a NGINX server.
I created a [custom theme](https://github.com/crivotz/keeweb-base16-default-light) that I set as default when I start the app (I created a theme folder inside keeweb).
It works perfectly the first time you access it, next time or when you refresh it, the theme is not loaded and is set to flat blue.

```
{
  "settings": {
    "theme": "base16-default-light",
    "idleMinutes": 10,
    "expandGroups": true,
    "menuViewWidth": 300,
    "listViewWidth": 340,
    "locale": "en",
    "colorfulIcons": true,
    "skipOpenLocalWarn": false,
    "skipHttpsWarning": false,
    "demoOpened": false,
    "fontSize": 1,
    "canOpen": false,
    "canOpenDemo": false,
    "canOpenSettings": false,
    "canCreate": false,
    "canImportXml": false,
    "webdav": false,
    "dropbox": false,
    "onedrive": false,
    "gdrive": false,
    "gdriveClientId": "client-id"
  },
  "files": [
    {
      "storage": "gdrive",
      "name": "first",
      "path": "first-file-id"
    },
    {
      "storage": "gdrive",
      "name": "second",
      "path": "second-file-id"
    }
  ],
  "plugins": [
    {
      "url": "themes/keeweb-base16-default-light"
    }
  ],
  "showOnlyFilesFromConfig": true
}
```

Here the devtool log after refresh/re-enter, it's seem a local cache problem

```
VM1093:1 Document was loaded from Application Cache with manifest https://abcabc/manifest.appcache
VM1093:1 Application Cache Checking event
VM1093:1 Application Cache NoUpdate event
(index):1 2018-11-19T15:59:08.512Z [plugin:keeweb-theme-base16-default-light] Manifest validation error Public key mismatch
a.error @ (index):1
(anonymous) @ (index):1
Promise.then (async)
install @ (index):1
loadPlugin @ (index):1
(anonymous) @ (index):1
(anonymous) @ (index):1
Promise.then (async)
(anonymous) @ (index):1
Promise.then (async)
init @ (index):1
C @ (index):1
Promise.then (async)
(anonymous) @ (index):1
j @ (index):1
k @ (index):1
setTimeout (async)
(anonymous) @ (index):1
i @ (index):1
fireWith @ (index):1
fire @ (index):1
i @ (index):1
fireWith @ (index):1
ready @ (index):1
[...]

PR #1051, opened 2018-12-18
Fixing issue where URL loaded plugins fail to automatically load after program restart

thecount2a:
Fixing issue where plugin manager does not persist any info across program restarts about whether a particular plugin was a user-installed plugin (and therefore is exempt from signature checking). This will allow URL installed plugins to stay across restarts of the program.

### DIFF
--- a/app/scripts/plugins/plugin-manager.js
+++ b/app/scripts/plugins/plugin-manager.js
@@ -173,13 +173,14 @@ const PluginManager = Backbone.Model.extend({
         const plugin = new Plugin({
             manifest: desc.manifest,
             url: desc.url,
-            autoUpdate: desc.autoUpdate
+            autoUpdate: desc.autoUpdate,
+            skipSignatureValidation: desc.skipSignatureValidation
         });
         let enabled = desc.enabled;
         if (enabled) {
             const galleryPlugin = gallery ? gallery.plugins.find(pl => pl.manifest.name === desc.manifest.name) : null;
             const expectedPublicKey = galleryPlugin ? galleryPlugin.manifest.publicKey : SignatureVerifier.getPublicKey();
-            enabled = desc.manifest.publicKey === expectedPublicKey;
+            enabled = (desc.manifest.publicKey === expectedPublicKey) || desc.skipSignatureValidation;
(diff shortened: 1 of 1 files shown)


antelle:
Thanks for contribution! 🎁 

However, I have to say, I'll close this.

This flag is not persistent, that's by design. The list of plugins, as well as other settings, are not protected and any other app can modify keeweb to include a malicious plugin. That's why we check signatures after start and allow to install custom plugins only if the application is self-hosted and the plugin is present in the config. This allows adding plugins into custom-built installations, without compromising security.
For desktop there should be a solution when additional keys are stored inside our write-protected application folder, it's not there now.

#1042 should be solved in another way: the flag [is passed](https://github.com/keeweb/keeweb/blob/master/app/scripts/models/app-model.js#L127) for plugins loaded from config, but for some reason they're not loaded, I haven't debugged it yet. That's what we need to fix to solve this issue.

thecount2a:
Okay, this kinda makes sense to me but it would be nice if this was better documented in the plugin building docs so people don't go through the work on developing a plugin just to realize that their plugin will never persist across reboots within the desktop apps. I thought developing plugins was supposed to be a quicker way to get additional functionality without rebuilding the entire project from scratch. If I must rebuild the entire project from scratch, just to get my generated key or config into the build, it doesn't actually save any time or effort. That being said, I understand the security aspect and I agree with the philosophy. I just wish I had known that there was a difference between a plugin being temporarily loaded (I guess for testing purposes only?) and actually persisting across restarts.

antelle:
The difference between temporary loading and persisting:
- temporary loaded plugins can be initiated only by user from the UI
- persistent plugins can be loaded by the application itself, based on values from the config. The lack of a secure way to store configs is exactly what is preventing us from having persistent plugins. For example, if we store plugins in the system keychain (or equivalent), there would be no objections regarding unsigned plugins. For now, any app can modify keeweb settings, so we don't trust values there and validate everything on start.

If you submit the plugin to https://github.com/keeweb/keeweb-plugins, it can be signed and it will be possible to install it permanently from https://plugins.keeweb.info.

PR #1052, opened 2018-12-21
Added name field. fixes #975

### LINKED ISSUE #975: [Feature Request] Name the password field
Hi,

I'm using a complex password to use Keepass, so when I'm on the login page of Keeweb I'd like my Lastpass extension to fill the password field, but it can do it because the password field seems to missing some requirements like the html `name` property.

Could you add it ? Or I'm I missing something?

dploeger:


antelle:
Not sure if it fixes the issue, but it won't harm. 🎁

PR #1053, opened 2018-12-21
Set file to be modified after deleting a version. Fixes #974

### LINKED ISSUE #974: Deleting versions in history is not saved
I have a PW entry with 10 versions. I want to keep no10 and delete (intermediate) versions (before my last good no10) no7, 8 and 9.
I deleted versions 7, 8 and 9. History shows me correctly that I now have 6 versions (besides the active one) left.
However when I close KeeWeb and reopen, all versions are back: 10 items.
I tried again and checked the filename in the bottom left corner: I'd expect it to get an asterisk as soon as I have an unsaved edit - but after deleting a version, it doesn't indicate any unsaved edit.
After I deleted some versions (again) I explicitly saved the db (CTRL+S) and closed it - still no difference: changes in history were not saved.

Note that when I *restore* a previous version, then the "History" entry (dimmed text to the right after retturning to the entry) indicates "changed" and the filename has its asterisk OK. So restoring an entry seems to be detected OK, but deleting entries goes unnoted.
BTW: Deleting versions works as expected in KeePass (unsaved edit is also indicated OK).

===========================
KeeWeb v1.6.3 (cded8a4, 2017-12-11)
Environment: electron v1.7.9
User-Agent: Mozilla/5.0 (Windows NT 6.3; Win64; x64) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.6.3 Chrome/58.0.3029.110 Electron/1.7.9 Safari/537.36

dploeger:


antelle:
thanks! 🎁

PR #1055, opened 2018-12-28
Select from last open files by using arrow keys. Fixes #889

### LINKED ISSUE #889: feature request: keyboard navigation/shortcuts for open dialog
I currently work with two .kdbx files, one for private stuff, one for my business related secrets. When opening (ctrl-o) a database file, the last opened .kdbx is still there, and the passphrase can be entered without any more actions/clicks. But then, opening the second one involves mouse clicking it. The same could be probably be achieved by a (tab [focus latest items list] => arrow-down [scroll one item down]) key sequence. I can guess that tab is already implemented (every third tab pressed focuses the passphrase field, again), but it's not clear how and what it does?

clearly optimizing the nifty small things - but that's why I really dig keeweb!

dploeger:


antelle:
Thanks! 🎁

PR #949, opened 2018-06-27
Migrate to webpack 4

twang2218:
Grunt => Webpack
=========================

* Extract webpack configuration to separated file.
    * `webpackConfig` => `webpack.config.app.js`: Webpack configuration for App content;
    * `webpack.config.main.js`: Webpack configuration for Electron main process;
        * Copy all files under `desktop/` to `dist/app/` except `package-lock.json;
        * Replace the `@@PUBLIC_KEY_CONTENT` with `app/resources/public-key.pem`'s content;
* Replace `replaceFont()` function with webpack `url-loader`;
* Rearrange `dist`, `tmp` directories to single `dist` directory:
    * `tmp`, `tmp/desktop/app`, `tmp/cordova` => `dist/app`: App content;
    * `dist`, `tmp/desktop` => `dist/build`: Build directory for each platform;
    * `dist/desktop/` => `dist/release`: Release files for each platform;
* `grunt --beta` => `webpack --env.release=beta`;
* Targets:
    * Web: `webpack --env.target=web`
    * Desktop: `webpack --env.target=desktop`
    * Cordova: `webpack --env.target=cordova`
* Release Channels:
    * Stable: `webpack --env.release=stable`
    * Beta: `webpack --env.release=beta`
* Production vs Development:
    * Production: `NODE_ENV=production`
    * Development: `NODE_ENV=development`
* Grunt Tasks:
    * `gitinfo`: `git-revision-webpack-plugin`;
    * `clean`: NPM script with `rimraf` package;
    * `copy`:
        * `copy:html`:
            * `app/index.html` => `app/index.hbs` template;
            * `html-webpack-plugin`;
        * `copy:favicon` => `html-webpack-plugin`;
        * `copy:icons` => `copy-webpack-plugin`;
        * `copy:dist-icons` => `copy-webpack-plugin`;
        * `copy:manifest` => `copy-webpack-plugin`;
        * `copy:dist-manifest` => `copy-webpack-plugin`;
        * `copy:fonts` => webpack's `url-loader`;
        * `copy:desktop-app-content` => `webpack.config.main.js`'s `copy-webpack-plugin`;
    * `eslint` => webpack `eslint-loader`;
    * `sass` => webpack `sass-loader`;
    * `postcss` => webpack plugins:
        * `resolve-url-loader`
        * `css-loader`
        * `optimize-css-assets-webpack-plugin`
        * `style-loader` (fallback)
    * `inline` => webpack plugins:
        * `script-ext-html-webpack-plugin`
[...]

PR #1056, opened 2018-12-28
Use non-minified versions of depending libraries on dev-server

dploeger:
I just modified the webpack config for the dev-server to use non-minified versions of libraries to allow for better debugging and thought I'd share. 😉

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -132,6 +132,23 @@ module.exports = function(grunt) {
         }
     };
 
+    let webpackDevConfig = webpackConfig
+    webpackDevConfig.resolve.alias = {
+        backbone: 'backbone/backbone.js',
+        underscore: 'underscore/underscore.js',
+        _: 'underscore/underscore.js',
+        jquery: 'jquery/dist/jquery.js',
+        kdbxweb: 'kdbxweb/dist/kdbxweb.js',
+        baron: 'baron/baron.js',
+        pikaday: 'pikaday/pikaday.js',
+        filesaver: 'FileSaver.js/FileSaver.js',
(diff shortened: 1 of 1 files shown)


Gruntfile.js
```diff
@@ -132,6 +132,23 @@ module.exports = function(grunt) {
         }
     };
 
+    let webpackDevConfig = webpackConfig
+    webpackDevConfig.resolve.alias = {
```
antelle: Probably you meant 
```js
Object.assign({}, webpackConfig, { resolve: ... });
```
This will `webpackConfig` now.

antelle:
Hi! 🎁  Thanks!
Added a comment there. For v1.7 there's also another PR about WebPack 4 migration: https://github.com/keeweb/keeweb/pull/949

Gruntfile.js
dploeger: Whoops, sorry. Yes. Corrected.

dploeger:
I will comment on #949 and ask the author to check this feature as well as my knowledge of webpack is failing here. 😉

Gruntfile.js
```diff
@@ -132,6 +132,25 @@ module.exports = function(grunt) {
         }
     };
 
+    let webpackDevConfig = Object.assign({}, webpackConfig, {
+        resolve:
+            alias = {
```
antelle: Not sure how this worked: the syntax is invalid, I've fixed it in the next commit after merge.

dploeger: Argh, and it still doesn't work, because I missed the modules key inside resolve and Object.assign just overwrites the resolve key. Will send a PR for that.

PR #1057, opened 2018-12-28
Use "Sync" instead of "Save" on cloud storage. Fixes #880

### LINKED ISSUE #880: Dropbox sync and "Save" button
The database is located in `Drobpox > Apps > KeeWeb` folder. Two-step verification on Dropbox is enabled. The database is used from Win7, Win10 and Android (via KeePass2Android). 

I already had a problem in the past (not more than two weeks ago, on the latest keeweb version) with local database on my PC being not refreshed. I had no time to create an issue at that moment, I solved the problem by deleting the complete Keeweb data folder altogether (on windows it is like %USERPROFILE%\AppData\Roaming\KeeWeb\).

On the https://keeweb.info/ I see the button "Sync with Dropbox", but my version has "Save", which is... a completely different thing IMHO.

I would like to clarify which button text is correct and in case if it depends on the type of the used storage (or whatever), what can be written on the button?

- app version: v1.6.3

dploeger:


app/templates/settings/settings-file.hbs
```diff
@@ -11,7 +11,10 @@
     {{/if}}
 
     <div class="settings__file-buttons">
-        <button class="settings__file-button-save-default" {{#if syncing}}disabled{{/if}}>{{res 'setFileSave'}}</button>
+        <button class="settings__file-button-save-default" {{#if syncing}}disabled{{/if}}>
+            {{#ifeq storage 'file'}}{{res 'setFileSave'}}{{/ifeq}}
+            {{#ifneq storage 'file'}}{{res 'setFileSync'}}{{/ifneq}}
```
antelle: `setFileSync` is a noun, not a verb. Let's add a new variable for this.

dploeger: I couldn't come up with a better name than "setFileSyncVerb".

antelle:
Could be "action" or something else. But "verb" should be also ok, at least it explains what it is. Thanks!

Issue #231, opened 2016-05-17
Persist advanced search options 


philogit:
Revisiting this topic (#222), I noticed a couple of possible related enhancements for how custom choices within advanced search options are handled:

1- **Ignore window state**: User choices in the advanced search options window are applied to the search results only when the window is open. Should custom choices continue to be applied when the advanced search options box is closed?

2- **Persist custom choices**: User choices in the advanced search options window are reverted to default when a file is closed and reopened. Should a user's custom choices be persisted between sessions?

===
KeeWeb v1.1.4 (8788e67, 2016-05-15)
Environment: web
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10.11; rv:46.0) Gecko/20100101 Firefox/46.0

antelle:
we can persist them, as now we're changing advanced search icon. so, it may change a bit, to indicate that settings are applied

ik9999:
@antelle Any updates on it?

ik9999:
@antelle Would it be ok to move the 
https://github.com/keeweb/keeweb/blob/master/app/scripts/views/list-search-view.js#L53
`this.advancedSearch` to settings json?
This will not persist them, but will allow the owner to customize the detfault filters.

antelle:
It would be still strange if other search settings are applied without UI displaying that.

ik9999:
@antelle What do you mean? There is a UI for displaying that.
![screenshot from 2018-12-29 20-07-38](https://user-images.githubusercontent.com/6804575/50540590-77dd1e00-0ba5-11e9-88f7-f1bf4503e600.png)

What i meant is to allow user to set the default filters instead of hard-coding them.
For example id like to only search the title. To avoid opening filters and deseleting other fields than title every time before searching, i would like to specify it in json app config.

antelle:
This won't be visible until advanced search settings are opened.
But perhaps, if it exists only in settings.json and is not saved automatically, it's quote ok to have it there, I agree. However it's not the same as was requested in the original issue.

antelle:
This makes sense.

> trigger add-filter with default advancedSearch on ListSearchView initialize

The only thing, it should be `false` by default, if there's no config option. Let's not set it to the default version if there's no config, because advanced search is a bit slower compared to normal search.

> app-model.js#L32

Probably what you mean is [here](https://github.com/keeweb/keeweb/blob/master/app/scripts/models/app-model.js#L107), when settings from the config are applied.

[...]

PR #1061, opened 2018-12-30
Allow custom default advancedSearch filters in settings json

ik9999:
In [this issue](https://github.com/keeweb/keeweb/issues/231) we had a discussion with @antelle about the ability customize default advancedSearch in json settings.
I've implemented this feature.

Add advancedSearch to your settings.json like this
```
{
  "settings": {
   ...
  },
  "advancedSearch": {
    "user": false,
    "other": true,
    "url": true,
    "protect": false,
    "notes": true,
    "pass": false,
    "cs": false,
    "regex": false,
    "history": false,
    "title": true
  },
  "showOnlyFilesFromConfig": false
}
```
and this filters will be activated by default instead of hard-coded ones.


antelle:
🎁

PR #1063, opened 2018-12-31
Fix: search filters UI opens when user clicks on folder with adancedSearch set in config.json

ik9999:
Related to ["Allow custom default advancedSearch filters in settings json" pull request](https://github.com/keeweb/keeweb/pull/1061).


antelle:
What is the issue that is solved here? Looks like search settings should not be modified from this place.

antelle:
I see, makes sense, thanks for the explanation!

PR #1067, opened 2019-01-02
Check for a URL in the title for unsupported apps

### LINKED ISSUE #812: Auto-type not finding entries in Waterfox
I'm using macOS 10.13.2 and global auto type works as expected in Safari. However, in Waterfox 56.0.1 (I assume Firefox as well, but haven't tried), it isn't finding entries automatically.

I tried using an addon which adds the URL to the window title as well, and there are still no results. The window appears, but with no related matches - I have to type in the the site name to get the entry.

dploeger:
This fixes #812, but I only included a Darwin implementation. What about the other platforms? I could easily move the regexp-tester to a mixin and use that in the other helpers. However, in the issue you spoke about implementing this only in the "non-supported apps" to avoid malicious things.


dploeger:
Done, but it's not for Darwin now anymore.

PR #1068, opened 2019-01-02
Remove a custom field from a template, better drag and drop visibility when moving folders

### LINKED ISSUE #791: Delete field in a template
Hi,
I use keeweb since today and I need a lot of different forms like in 1password from which I came.
For this I have to enter a lot of different forms as templates. On working on it the following question comes up: How can I delete a field in a template?

Another thing is what i see is that templates are shown as a normal subfolder. From a infomation structure view i would say that the templates should be outside of the normal folders.
But this is only a thought spark :).

Cheers

dploeger:
Fixes #791 for both mentioned issues there.

I implemented removing a custom field by emptying its title. Currently, this was only possible by emptying its value, which broke templates support. However, this introduces some kind of breaking change. Should I narrow this implementation so it only works for templates?

### DIFF
--- a/app/scripts/views/fields/field-view-custom.js
+++ b/app/scripts/views/fields/field-view-custom.js
@@ -68,6 +68,11 @@ const FieldViewCustom = FieldViewText.extend({
         if (newTitle && newTitle !== this.model.title) {
             this.model.title = newTitle;
             this.model.titleChanged = true;
+        } else if (newTitle === '') {
+            this.trigger('change', {
+                field: '$' + this.model.title,
+                val: ''
+            });
         }
         this.$el.find('.details__field-label').text(this.model.title);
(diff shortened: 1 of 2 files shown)


antelle:
> Should I narrow this implementation so it only works for templates?

Yes, now it will break custom fields editing, if I'm not mistaken.

dploeger:
> Yes, now it will break custom fields editing, if I'm not mistaken.

No, editing works as before. You're just allowed to empty the title of a field and this will remove the field. Previously, clearing the title would reset the title to the original value.

So "break" is a big word for it, but some users still could be irritated, right?

antelle:
Thanks! Merged it, actually it doesn't seem to be annoying, let's see if anyone reports it after the release.

PR #1074, opened 2019-01-04
Missing modules. Fixes a bug introduced in #1056

dploeger:


Gruntfile.js
```diff
@@ -134,6 +134,7 @@ module.exports = function(grunt) {
 
     const webpackDevConfig = Object.assign({}, webpackConfig, {
         resolve: {
+            modules: [path.join(__dirname, 'app/scripts'), path.join(__dirname, 'node_modules')],
```
antelle: Since this line is equal, probably it's better to assign only alias to a new object, to avoid repeating this code here.

antelle:
fixed this in 152bea3547076939297e29f58d6f18a0005763f6

PR #1088, opened 2019-01-13
Only focuses input field when app or browser is active on lock

### LINKED ISSUE #423: Sometimes Keeweb prevents Alfred auto-expansion feature from working
Suddenly the auto-expansion feature of my [Alfred](//alfredapp.com) stopped working. I researched the forums and found this:

> It's likely that you're experiencing the same issue as other users where secure input isn't released correctly by a plugin or app that requires it, and restarting your Mac simply reset the secure input. Next time, instead of rebooting, try quitting browsers to see whether it's a plugin like LastPass at the source of the issue. — from [here](http://www.alfredforum.com/topic/8871-alfred-3-snippets-dont-auto-expand-fix-awaiting-chromium-bug-fix/?page=2)

I started exiting apps one by one and checking whether such and such app was the culprit. Finally I was able to figure out that restarting KeeWeb fixes the problem. Sorry I cannot provide more details as I am unfamiliar with how OS X works in this regard.

But maybe you can check and find the bug without additional information, using just this description. By the way I haven’t even used KeeWeb this time, it just loaded automatically on boot.

dploeger:
Fixes #423

Should be included in the next patch release as it is a major usability issue especially on macOS.


app/scripts/views/open-view.js
```diff
@@ -107,7 +111,12 @@ const OpenView = Backbone.View.extend({
 
     focusInput: function() {
         if (!FeatureDetector.isMobile) {
-            this.inputEl.focus();
+            if (
+                (FeatureDetector.isDesktop && Launcher.isAppFocused()) ||
```
antelle: Let's put web/desktop logic to the component and return a value from isFocused. Looks like conceptually there's no difference between web and desktop, we just get the property in a different way.

dploeger: What component do you mean? The launcher (there currently is no web launcher, right?) or the open view?

antelle: "browser-focus", that was added here :)

dploeger: Ah, allright. Done that. 😀

app/scripts/util/focus-detector.js
```diff
@@ -0,0 +1,20 @@
+const FeatureDetector = require('./feature-detector');
+const Launcher = require('../comp/launcher');
+
+const FocusDetector = function () {
+    this.isFocused = false;
+    if (FeatureDetector.isBrowser) {
```
antelle: I would expect `isBrowser` to be true on mobile, but I'll fix this after merge.

app/scripts/util/focus-detector.js
```diff
@@ -0,0 +1,20 @@
+const FeatureDetector = require('./feature-detector');
+const Launcher = require('../comp/launcher');
+
+const FocusDetector = function () {
+    this.isFocused = false;
```
antelle: I don't think this default is the proper value, changed it to `true` because it broke initial focus when you have recent files. Please check if the fix is still working.

antelle:
Pushed some improvements in 903d6dc271f7f10542679d0fff942890e1e42456

PR #1103, opened 2019-02-05
Add close window item to menu on Mac

umlx5h:
I realized close window shortcut doesn't work on Mac, therefore I added this item to menu.


antelle:
Thank you! 🎁

PR #1107, opened 2019-02-07
Try to download favicon using proxy when fail to fetch on desktop app

### LINKED ISSUE #968: Improve favicon detection
[Currently](https://github.com/keeweb/keeweb/blob/abd8f6d20a6b61903090c9432063e9498507562b/app/scripts/views/icon-select-view.js#L84) only search for a favicon.ico url at root, which doesn't work for a lots of website.

Exemples of undetected favicon :

```html
<link rel="icon" type="image/png" href="XXX/favicon-32x32.png" sizes="32x32">
<link rel="shortcut icon" type="image/x-icon" href="XXX/logo_v10.png" />
<link rel="icon" type="image/x-icon" class="js-site-favicon" href="XXX/favicon.ico">
<!-- Below is in addition to icon tag but can sometime be better -->
<link rel="mask-icon" href="/favicon.svg" color="#990000">
```

umlx5h:
* FIxed #968

Hi.
I also made PR to keeweb/favicon-proxy#5 in order to improve getting favicon method.
I can't download favicon in some sites from desktop app. So I changed the downloading behavior as follows.
First, download favicon as usual and if it fails, try to download favicon via proxy server only when client is desktop app.
I confirm the operation from Mac & Windows applications and web.

### DIFF
--- a/app/scripts/views/icon-select-view.js
+++ b/app/scripts/views/icon-select-view.js
@@ -49,14 +49,16 @@ const IconSelectView = Backbone.View.extend({
         }
     },
 
-    downloadIcon: function() {
-        if (this.downloadingFavicon) {
+    downloadIcon: function(e, isSecondO) {
+        const isSecond = isSecondO || false;
+
+        if (isSecond === false && this.downloadingFavicon) {
             return;
         }
         this.downloadingFavicon = true;
         this.$el.find('.icon-select__icon-download>i').addClass('fa-spinner fa-spin');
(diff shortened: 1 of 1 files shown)


antelle:
Hi! Thanks! 🎁 
If a desktop app can go to favicon.keeweb.info in some cases, it should do it always. Changed it in f729872d70dd32848abc0f3c6f0b3a8167adf2e6

PR #1120, opened 2019-02-22
dev-env.sh: allow to clone via http url

return42:
BTW, instead of cloning keeweb twice add a git-worktree for the gh-pages.

Signed-off-by: Markus Heiser <markus.heiser@darmarit.de>

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ To make Dropbox work in your self-hosted app, [go to this Wiki page](https://git
 
 The easiest way to clone all KeeWeb repos is:
 ```bash
-curl https://raw.githubusercontent.com/keeweb/keeweb/develop/dev-env.sh | bash -
+curl https://raw.githubusercontent.com/keeweb/keeweb/develop/dev-env.sh | SCHEMA=http bash -
 ```
 
 The app can be built with grunt: `grunt` (html file will be in `dist/`).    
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -37,7 +37,7 @@ To make Dropbox work in your self-hosted app, [go to this Wiki page](https://git
 
 The easiest way to clone all KeeWeb repos is:
 ```bash
-curl https://raw.githubusercontent.com/keeweb/keeweb/develop/dev-env.sh | bash -
+curl https://raw.githubusercontent.com/keeweb/keeweb/develop/dev-env.sh | SCHEMA=http bash -
```
antelle: Let's leave the default option here.

antelle:
Thank you! 🎁

README.md
return42: OK, done.. you know that _other_ user get ... _could not read_ .. ?
````
Cloning into 'keeweb'...
git@github.com: Permission denied (publickey).
fatal: Could not read from remote repository.

Please make sure you have the correct access rights
and the repository exists.
````

antelle: Then we can make the https option default in the script, if we think it's what will work for the most of users.

return42: Should I patch? (I won't change the default behavior, because I don't know if this is used by a build bot)

antelle: Sure! This script is only for humans who want to clone all the repos.

PR #1121, opened 2019-02-22
dev-env.sh: use HTTP schema as default for cloning

return42:
It's what will work for the most of users.

Suggested-by: @antelle
Signed-off-by: Markus Heiser <markus.heiser@darmarit.de>


antelle:
Makes sense, thank you! 🎁

PR #1122, opened 2019-02-22
icon-select-view.js: remove unused Launcher (eslint:app)

return42:
Is it only me getting this error? I'am using eslint 5.14.1 (linux)


antelle:
Same here indeed, thank you 🎁

PR #1123, opened 2019-02-22
Focus search field when showing view

etienne-martin:
Automatically focus on search input whenever a user lands on the main page. Won't work on iOS.


antelle:
I don't think we need it. The search starts working when the interface is shown, we don't need to focus the field and draw user's attention here.

etienne-martin:
I just realized that we can tab into it, fair enough.

antelle:
an tab as well, you're right 😉

PR #1126, opened 2019-03-03
New Feature: Disable direct autotype

dploeger:
In this PR I implemented a setting to disable the direct run of Autotype if one specific entry can be found.

This is something I need in connection to the feature of specifically entering only the password or username using Autotype.

The default value of this setting is *true*, which mimics the current behavior, so this is backwards compatible and should be valid for a minor version bump.

### DIFF
--- a/app/scripts/auto-type/index.js
+++ b/app/scripts/auto-type/index.js
@@ -199,7 +199,7 @@ const AutoType = {
 
     processEventWithFilter(evt) {
         const entries = evt.filter.getEntries();
-        if (entries.length === 1) {
+        if (entries.length === 1 && AppSettingsModel.instance.get('directAutotype')) {
             this.hideWindow(() => {
                 this.runAndHandleResult({ entry: entries.at(0) });
             });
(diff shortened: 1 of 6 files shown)


antelle:
Hi! Thank 🎁 
Let's change the name of the checkbox, I wouldn't understand what it means. Perhaps should be called something "Directly run auto-type if only one matching entry is found".

dploeger:
Yeah, I was unsure about that as well. Check out my new suggestion.

antelle:
It looks better now!  
Let's hide the setting for web, i.e. where auto-type is not supported. Also please move it to "Function", it doesn't seem to be something general.

dploeger:
You're right. Moved it to function.

antelle:
Thank you! 🎁

dploeger:
@antelle For what release do you plan to include this feature?

antelle:
should be 1.8

dploeger:
Great. 1.8 is released when [these issues](https://github.com/keeweb/keeweb/issues?q=is%3Aopen+is%3Aissue+milestone%3Av1.8+no%3Aassignee) are fixed?

antelle:
I haven’t planned it yet, it’s just a quick list of possible issues

dploeger:
Okay, I try to fix some of those if I find the time, because I need this feature.

antelle:
Probably it will take some time to fix them. In theory, we can include this into other hotfixes, or make a new version with auto-type improvements only.

dploeger:
I like that theory. 😆

dploeger:
Maybe include #1128, too.

PR #1128, opened 2019-03-04
New feature: Open File from Autotype

dploeger:
This PR adds a shortcut to open a new database from the autotype screen. This somehow completes my workflow on working with multiple databases. 😄

### DIFF
--- a/app/scripts/auto-type/index.js
+++ b/app/scripts/auto-type/index.js
@@ -221,6 +221,14 @@ const AutoType = {
                 }
             });
         });
+        this.selectEntryView.on('openFiles', () => {
+            this.selectEntryView.hide();
+            Backbone.trigger('open-file');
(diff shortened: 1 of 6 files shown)


antelle:
Thanks, merged in https://github.com/keeweb/keeweb/commit/35ed03021a8f0a7903d9594be395772d8f62ca79 (also made some changes there)

dploeger:
@antelle Hm. This seems to work in 1.8, however the hint text is missing on the autotype view.

dploeger:
Thanks btw. 😄

antelle:
You mean the shortcut description? I've removed it intentionally, the shortcut should "just work" as it does on other screens. While other shortcuts are special to auto-type, so we display a hint about it.

dploeger:
Ah, okay. Got it.

PR #1131, opened 2019-03-07
Lowercased the search string, so case doesn't produce different ranking results

dploeger:
Fixes the issue, that came up in #1073 after merging

### DIFF
--- a/app/scripts/models/entry-model.js
+++ b/app/scripts/models/entry-model.js
@@ -670,7 +670,7 @@ const EntryModel = Backbone.Model.extend({
         _.forEach(ranking, rankingEntry => {
             if (this._getFieldString(rankingEntry.field).toLowerCase() !== '') {
                 const calculatedRank = Ranking.getStringRank(
-                    searchString,
+                    searchString.toLowerCase(),
                     this._getFieldString(rankingEntry.field).toLowerCase()
                 ) * rankingEntry.multiplicator;
                 rank += calculatedRank;
--- a/app/scripts/models/entry-model.js
+++ b/app/scripts/models/entry-model.js
@@ -670,7 +670,7 @@ const EntryModel = Backbone.Model.extend({
         _.forEach(ranking, rankingEntry => {
             if (this._getFieldString(rankingEntry.field).toLowerCase() !== '') {
                 const calculatedRank = Ranking.getStringRank(
-                    searchString.toLowerCase(),
+                    searchString,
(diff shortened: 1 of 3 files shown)


app/scripts/models/entry-model.js
```diff
@@ -670,7 +670,7 @@ const EntryModel = Backbone.Model.extend({
         _.forEach(ranking, rankingEntry => {
             if (this._getFieldString(rankingEntry.field).toLowerCase() !== '') {
                 const calculatedRank = Ranking.getStringRank(
-                    searchString,
+                    searchString.toLowerCase(),
```
antelle: Let's pass `textLower` in `rankComparator`, instead of doing `toLowerCase` for every entry.

dploeger: Hmm... textLower doesn't reach the rankComparator. It is prepared here:

[https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L285](https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L285)

but the prepared filter including textLower is not persisted into this.filter, so it is lost again here:

[https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L277](https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L277)

Shouldn't the filter better be prepared here:
[https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L267](https://github.com/keeweb/keeweb/blob/f0ff26ebb4e40f4c8728230b585f052feb471078/app/scripts/models/app-model.js#L267)

antelle: Hm, isn't it passed here? https://github.com/keeweb/keeweb/blob/develop/app/scripts/util/comparators.js#L16

antelle: Read it again, I see. Yes, I think it makes sense to move it there

dploeger:
Done

app/scripts/models/entry-model.js
dploeger: Done

antelle:
Thanks! 🎁

PR #1134, opened 2019-03-12
Prevent default on details copy keyboard shortcuts

luckyrat:
This fixes a bug in Chrome where the "View Source" feature is invoked when attempting to copy the URL value via Ctrl+U (at least on Linux Chromium).

I've applied the same change to C and B for consistency and future-proofing, although I see no current bugs caused by the existing behaviour of allowing the events to attempt their default actions (presumably they are a noop in most browsers since no text is selected at the time).

Tested in browsers but not electron.


antelle:
Thanks! 🎁

PR #1148, opened 2019-03-25
Change string comparator last char value to 0xfffd

### LINKED ISSUE #1147: Empty titles breaks KeeWeb in MS Edge
Symptoms in MS Edge (v17) include being unable to open any file containing an entry with an empty title and triggering what appear to be fatal bugs when creating new entries (I've not found a way to work around yet but there's maybe a small chance the bugs are limited to failed rendering of the new entries rather than affecting data integrity). In any case, this makes usage of KeeWeb in MS Edge unreliable.

Quickest way to reproduce is to load the demo database and add two entries one after another, without entering any contents for the title of the first entry. From then on the UI doesn't work properly (missing items, incorrect ordering, unresponsive clicking on some items). I haven't dug into exactly what bits of the UI break or exactly why because the underlying bug is indicated in the console:

`Failed to normalize: invalid or missing unicode character at index 0`

To fix, we need to change the string comparison "maximum character value" to `\ufffd`

The current value of `\uffffffff` per the spec of fromCharCode gets truncated to `\uffff`

`\uffff` (and `\ufffe`) are "Invalid Character"s and Edge throws an exception when making any attempt to compare against these characters.

I'll make a PR in a minute which contains the same fix I've used and tested in Kee Vault.

I note the comment "`// bugged in Edge: #808`" which suggests that in the past it was sufficient to bypass the Intl.Collator function so perhaps future Edge development has aligned the localeCompare function with the same Intl.Collator implementation (although I can find no details about this possible bug and the original reporter never confirmed that issue #808 was fixed).

For Kee Vault I have removed that Edge-specific code path and it works for me but it's possible older versions of Edge than I'm interested in supporting might still benefit from that workaround. I'll report back if I hear of any real world issues in old versions but if not you might want to simplify that eventually too.

luckyrat:
Fixes #1147 

The only risks I see are:

1) If a user has an "Invalid Character" in their titles (or other strings compared by this function) the sort order will change.
2) If other JS runtimes perform a comparison using UTF32 characters.

(1) Seems irrelevant; Unlikely such characters have any benefit for users and may cause a whole host of other problems in KeeWeb other kdbx-based apps anyway.
(2) Seems unlikely since the fromCharCode spec only accepts UTF16 characters so any deviation is likely to be a significant bug in those runtimes. At least the behaviour of `0xffff` is identical to `0xffffffff` in Chromium and Firefox.

I've tested a fixed version (via Kee Vault) in Edge, Chrome and Firefox. Happy to test on my Windows machine using the KeeWeb beta site when the fix is merged, although Windows' update notifications suggest I'll probably be on a newer version of Edge by then.

### DIFF
--- a/app/scripts/util/comparators.js
+++ b/app/scripts/util/comparators.js
@@ -1,4 +1,4 @@
-const LastChar = String.fromCharCode(0xffffffff);
+const LastChar = String.fromCharCode(0xfffd);
 
 const ciCompare = (window.Intl && window.Intl.Collator && !/Edge/.test(navigator.userAgent)) // bugged in Edge: #808
     ? new Intl.Collator(undefined, { sensitivity: 'base' }).compare


antelle:
Thanks! 🎁

Issue #156, opened 2016-03-09
Use ServiceWorker when possible


antelle:
Updater should use ServiceWorker if it's available in browser, with fallback to appCache.
https://ponyfoo.com/articles/simple-offline-site-serviceworker
- [x] add app manifest
- [x] use service worker
- [x] investigate signature checking inside the worker

davinkevin:
Very old issue but still important from my point of view. Especially when using it on mobile device (android, ios) without any application, having the offline would be perfect !

antelle:
ServiceWorker is everywhere, so the next version will be already using it instead of AppCache.

PR #1174, opened 2019-05-05
Replace Electron with Desktop PWA

kinolaev:
Hello @antelle!

I made keeweb [desktop PWA](https://developers.google.com/web/progressive-web-apps/desktop). I made it by replacing all your grunt/webpack/electron stack with react-scripts from create-react-app (it provide good build system even if you don't use react).

This PR contains only changes that don't break current build system but required to work with CRA (see commit messages). You can find actual build system change in https://github.com/kinolaev/keeweb/tree/cra, clone it and run `npm i && npm run build && npm i -g serve && serve -s build`.

Pros:
- no separate chromium installation
- simplified build system

Cons:
- can't use os filesystem

Other:
- ability to use react and typescript

Implementation details:
- raw/base64/handlebars loaders replaced with simple babel macro (src/loader.macro).

Future:
- handle `beforeinstallprompt` event and add install button

Related:
- #156

What do you think about this?


antelle:
Hi!

#### Electron
I don't think we should replace Electron with something that doesn't work without installed Chrome.

#### React
Let's not migrate _parts of the application_ to react. There's a branch where I tried to do kind or work on react migration, but it's very far from being ready. Adding an ability to use react while the app is written on another framework is not clear to me. Having two frameworks will make the application messed up.

Sorry for all efforts spend on work here, but I'll close it.

kinolaev:
Please, give me one more chance to explain my intention)

I don't suggest to migrate right now, I just want suggest a direction. I think when Firefox and Safari will support desktop PWA it will be better distribution form for keeweb then Electron. CRA is just easiest way to make desktop PWA with ESnext, SASS, and macros, without configuration (IMHO). I don't suggest to migrate to react and I didn't add react to keeweb in this PR nor in branch cra. It'is up to you if you want it.

**I suggest to migrate to CRA (not react) when Desktop PWA will be widely supported. Until that I can provide separate branch (cra) for thoose who want to use Desktop PWA now**. This PR contains only non-breaking changes that makes support of separate cra branch simple.

Does it makes sense for you?

antelle:
> when Firefox and Safari will support desktop PWA

Then we'll most likely switch to it, if it's good enough compared to electron. I'm not sure when (if) it will happen though.

I don't see the point in migrating the current build system (webpack) to something else, it works quite well and solves the needs that we have now.

Maybe we'll switch when the desktop pwa is widely adopted, for now I can't say for sure. Question about this: why can't it built with webpack? I don't have an experience with desktop pwa's, is there a reason why it can't use the output of our existing build?

kinolaev:
Ok, I understand.

I tried to add [workbox](https://developers.google.com/web/tools/workbox/) (used by cra for pwa), but migration to CRA turned out to be easier for me and then I thought that maybe you will like it).

I will try again later. 

Thank you for great app!

antelle:
Good, so there's no direct dependency, thanks!

Let's see how Desktop PWA's develop in other browsers, maybe it will work well in future.

kinolaev:
Last question: did you see the diff? Actually it doesn't contains anything special for CRA, only "best practices" (IMHO):
1. use ES modules instead of CommonJS
2. `location` and `screen` prefixed with `window`
3. `$` and `_` imported explicitly
4. import `handlebars/runtime` instead of `hbs`
5. `app` folder renamed to `src`
6. index.html, manifests and icons moved to `public` folder

Just for notice, no need to answer anything if you don't interesting.

antelle:
1. this is already done in react branch, as well as stricter ESLint config, prettier and other improvements; for now, it won't be merged, but just a general consideration: if it was going to be merged, it should've been a separate pull request containing only these files, otherwise it makes it impossible to review other changes
2. controversial
3. they're obsolete libraries; apart from this, these are imported by webpack, I don't see any harm in doing this implicitly
4. alias doesn't makes it worse; controversial
5. controversial
6. controversial

Issue #1215, opened 2019-07-23
keewebhttp should match fields even if no url is specified


asbachb:
### Summary
Currently the ranking algorithm which is invoked by keewebhttp does not handle entries when no "Website" entry was defined.

### Use Case
In case of an SSO environment for me it seems more applicable to define these environments as additional fields
e.g.
* Application1: http://application1.local/login
* Application2: https://application2.local/

### Expectation
The ranking algorithm should not differ between "Website" fields and other fields. The "Website" field might get a higher ranking than other fields. But they should not be completely ignored when not "Website" is defined.

### Links
Related Issues: #285

antelle:
Currently they're not ignored, but the url must match exactly. Doesn't it work for your setup?

asbachb:
> Currently they're not ignored, but the url must match exactly. Doesn't it work for your setup?

It works only when then "Website" field is not empty. With that I see several problems
a) The bevahiour is a little bit unclear. Why should a user expect that the feature only works when another field is filled?
b) If you're in a business environment when your windows account information is used for several web applications. What's your primary one? For me it feels more natural to define custom fields for these web applications.

antelle:
You mean, it works everywhere when you put something to website, right?  
In this case, you can put the url of the identity provider to the website field (i.e. where you change the password, create accounts, etc... your federation system), which will fix it.

asbachb:
> You mean, it works everywhere when you put something to website, right?
Yes.

> In this case, you can put the url of the identity provider to the website field (i.e. where you change the password, create accounts, etc... your federation system), which will fix it.
In case of a windows account there's no IDM.

Anyway I wonder if a functionality should be only available if some kind of field is filled. For a user this behavior (at least for me) does not seem to be obvious.

antelle:
It shouldn't of course, but now the issue becomes much easier to fix, I'll fix it.

antelle:
Done, next release will behave as expected.

PR #1216, opened 2019-07-23
keeweb#1215: Introduced ability to rank all fields independent of an …

asbachb:
**DISCLAIMER: I'm not a JavaScript developer. I'd highly appreciate a code review.**

Introduced ability to rank all fields independent of an set "Website" (`entry.url`) field.

All fields now use the same ranking algorithm. In case the same url is defined for "Website" and another field the "Website" field is ranked higher.

### DIFF
--- a/app/scripts/auto-type/auto-type-filter.js
+++ b/app/scripts/auto-type/auto-type-filter.js
@@ -1,6 +1,9 @@
 const EntryCollection = require('../collections/entry-collection');
 const Ranking = require('../util/ranking');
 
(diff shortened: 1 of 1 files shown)


asbachb:
@antelle - After your comment in #1215 I thought a little bit about that PR.

If this function is also used for auto type detection I guess it should not be merged since I just refactored it with keepass http in mind. I guess it might cause some trouble if it's also used for AutoType ranking.

antelle:
I'll close this one

Issue #1222, opened 2019-08-05
Option to hide password in generator


GrantMoyer:
It would handy to be able to hide the password in the generator. Not even knowing a password to a service is an extra layer of security.

The password generator dialog should have a toggle button which hides and shows the password, and the setting should persist across sessions.

Alternatives include closing your eyes while generating the password, or putting a "hide generated passwords" option somewhere in the settings. The first is impractical and I think the second hurts discoverability.

I have a PR ready to go if this is a desired feature; you can try it out on [my develop branch][devel]. My implementation looks like this:

[devel]: https://github.com/GrantMoyer/keeweb/tree/develop

|![Screen Shot 2019-08-05 at 19 23 33](https://user-images.githubusercontent.com/6333127/62501242-e5929d80-b7b7-11e9-9567-75275f4ca120.png) | ![Screen Shot 2019-08-05 at 19 24 18](https://user-images.githubusercontent.com/6333127/62501244-ea575180-b7b7-11e9-862e-0755df5629a7.png) |
|----|----|

GrantMoyer:
Some more screen caps for completeness:

| ![Short password visible][short-vis] | ![Short password hidden][short-hid] |
|----|----|
| ![Long password visible][long-vis] | ![Long password hidden][long-hid] |


[short-vis]: https://user-images.githubusercontent.com/6333127/62501717-12e04b00-b7ba-11e9-9e46-86e52fb42283.png
[short-hid]: https://user-images.githubusercontent.com/6333127/62501779-4f13ab80-b7ba-11e9-9fb5-7d1751ac192d.png
[long-vis]: https://user-images.githubusercontent.com/6333127/62501782-53d85f80-b7ba-11e9-91f3-0da346377dca.png
[long-hid]: https://user-images.githubusercontent.com/6333127/62501786-55a22300-b7ba-11e9-8d4b-8f8b041f5317.png

antelle:
Hi!

It makes sense, however the usage of this option seems to be somewhat limited, because the password is displayed anyway when you edit the password field. The only possible use case that I see is to paste the password and forget it forever, i.e. don't save it to the password manager, however it's strange. To complete the feature, we need to:

1. make it possible to show / hide passwords when editing
2. move the icon, it's not in the right place now (but that's much easier indeed)

GrantMoyer:
The current behavior is:

* If you copy and paste the password into the password field you can see the password, since you are editing the field.
[...]

PR #1223, opened 2019-08-07
Option to hide password in generator

GrantMoyer:
Implements  #1222

Demo:
demo

### DIFF
--- a/app/scripts/models/app-settings-model.js
+++ b/app/scripts/models/app-settings-model.js
@@ -33,7 +33,7 @@ const AppSettingsModel = Backbone.Model.extend({
         fontSize: 0,
         tableViewColumns: null,
         generatorPresets: null,
-        generatorHide: false,
+        generatorHidePassword: false,
         cacheConfigSettings: false,
 
         canOpen: true,
--- a/app/scripts/models/app-settings-model.js
+++ b/app/scripts/models/app-settings-model.js
@@ -33,6 +33,7 @@ const AppSettingsModel = Backbone.Model.extend({
         fontSize: 0,
         tableViewColumns: null,
         generatorPresets: null,
+        generatorHide: false,
         cacheConfigSettings: false,
 
         canOpen: true,
--- a/app/scripts/util/tip.js
+++ b/app/scripts/util/tip.js
@@ -11,6 +11,8 @@ const Tip = function(el, config) {
     this.hideTimeout = null;
     this.force = config && config.force || false;
     this.hide = this.hide.bind(this);
+    this.mouseenterHandler = null;
+    this.mouseleaveHandler = null;
 };
 
 Tip.enabled = !FeatureDetector.isMobile;
@@ -21,8 +23,10 @@ Tip.prototype.init = function() {
     }
     this.el.removeAttr('title');
     this.el.attr('data-title', this.title);
-    this.el.mouseenter(this.mouseenter.bind(this)).mouseleave(this.mouseleave.bind(this));
-    this.el.click(this.mouseleave.bind(this));
+    this.mouseenterHandler = this.mouseenter.bind(this);
+    this.mouseleaveHandler = this.mouseleave.bind(this);
+    this.el.mouseenter(this.mouseenterHandler).mouseleave(this.mouseleaveHandler);
+    this.el.click(this.mouseleaveHandler);
 };
 
 Tip.prototype.show = function() {
@@ -132,7 +136,7 @@ Tip.prototype.getAutoPlacement = function(rect, tipRect) {
     }
 };
 
-Tip.createTips = function(container) {
+Tip.createTips = function(container, options) {
     if (!Tip.enabled) {
         return;
     }
@@ -144,6 +148,12 @@ Tip.createTips = function(container) {
 Tip.createTip = function(el, options) {
     if (!Tip.enabled && (!options || !options.force)) {
         return;
+    }
+    if (el._tip) {
+        el._tip.el
+            .off('mouseenter', el._tip.mouseenterHandler)
+            .off('mouseleave', el._tip.mouseleaveHandler)
+            .off('click', el._tip.mouseleaveHandler);
     }
     const tip = new Tip($(el), options);
     if (!options || !options.noInit) {
--- a/app/scripts/util/tip.js
+++ b/app/scripts/util/tip.js
@@ -136,7 +136,7 @@ Tip.prototype.getAutoPlacement = function(rect, tipRect) {
     }
 };
 
-Tip.createTips = function(container, options) {
+Tip.createTips = function(container) {
     if (!Tip.enabled) {
         return;
     }
--- a/app/scripts/views/generator-view.js
+++ b/app/scripts/views/generator-view.js
@@ -1,6 +1,7 @@
 const Backbone = require('backbone');
 const PasswordGenerator = require('../util/password-generator');
 const CopyPaste = require('../comp/copy-paste');
+const AppSettingsModel = require('../models/app-settings-model');
 const GeneratorPresets = require('../comp/generator-presets');
 const Locale = require('../util/locale');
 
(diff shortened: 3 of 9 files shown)


GrantMoyer:
With 93b05e4, I think this now meets all the requirements we discussed, and is ready to merge.

app/scripts/views/generator-view.js
```diff
@@ -69,6 +80,15 @@ const GeneratorView = Backbone.View.extend({
         return this.valuesMap.length - 1;
     },
 
+    showPassword: function() {
+        if (this.hide && !this.model.copy) {
+            const dots = '•'.repeat(this.password.length);
```
antelle: `util/password-genertor::present`

app/scripts/models/app-settings-model.js
```diff
@@ -33,6 +33,7 @@ const AppSettingsModel = Backbone.Model.extend({
         fontSize: 0,
         tableViewColumns: null,
         generatorPresets: null,
+        generatorHide: false,
```
antelle: let's rename it to `generatorHidePassword` to make it more obvious

app/scripts/views/generator-view.js
```diff
@@ -30,14 +33,22 @@ const GeneratorView = Backbone.View.extend({
...
         this.listenTo(Backbone, 'lock-workspace', this.remove.bind(this));
     },
 
     render: function() {
         const canCopy = document.queryCommandSupported('copy');
         const btnTitle = this.model.copy ? canCopy ? Locale.alertCopy : Locale.alertClose : Locale.alertOk;
-        this.renderTemplate({ btnTitle: btnTitle, opt: this.gen, presets: this.presets, preset: this.preset });
+        this.renderTemplate({
+            btnTitle: btnTitle,
+            copy: this.model.copy,
```
antelle: let's provide it as `showToggleButton`

app/scripts/views/generator-view.js
```diff
@@ -103,17 +123,22 @@ const GeneratorView = Backbone.View.extend({
...
         this.password = PasswordGenerator.generate(this.gen);
-        this.resultEl.text(this.password);
+        this.showPassword();
         const isLong = this.password.length > 32;
         this.resultEl.toggleClass('gen__result--long-pass', isLong);
     },
 
+    hideChange: function(e) {
+        this.hide = e.target.checked;
+        AppSettingsModel.instance.unset('generatorHide', { silent: true });
```
antelle: Looks like it can be removed because we set it later.

app/scripts/util/tip.js
```diff
@@ -11,6 +11,8 @@ const Tip = function(el, config) {
     this.hideTimeout = null;
     this.force = config && config.force || false;
     this.hide = this.hide.bind(this);
+    this.mouseenterHandler = null;
```
antelle: Why do we need all this mouse magic instead of using the standard tooltips?

app/scripts/views/generator-view.js
```diff
@@ -103,17 +123,22 @@ const GeneratorView = Backbone.View.extend({
...
+        AppSettingsModel.instance.set('generatorHide', this.hide);
+        const label = this.$el.find('#gen__check-hide + label');
+        Tip.createTip(label[0], {title: this.hide ? Locale.genShowPass : Locale.genHidePass});
```
antelle: Let's remove this. Once you have pressed the button, you clearly see the change in the UI.

app/scripts/views/generator-view.js
```diff
@@ -103,17 +123,22 @@ const GeneratorView = Backbone.View.extend({
...
+        this.showPassword();
+    },
+
     btnOkClick: function() {
-        const selection = window.getSelection();
-        const range = document.createRange();
-        range.selectNodeContents(this.resultEl[0]);
-        selection.removeAllRanges();
-        selection.addRange(range);
+        CopyPaste.createHiddenInput(this.password);
```
antelle: Please also check `!CopyPaste.simpleCopy` before.

app/scripts/models/app-settings-model.js
GrantMoyer: Okay, makes sense.

app/scripts/util/tip.js
GrantMoyer: This was the system in place that's used for tge refresh button. I think we don't use native tooltips because you can't style them. Maybe it also has something to do with positioning.

app/scripts/views/generator-view.js
GrantMoyer: Makes sense.

app/scripts/views/generator-view.js
GrantMoyer: I'll switch it to using that function.

app/scripts/views/generator-view.js
GrantMoyer: I'm not sure what you mean. This is tge callback called when you press tge button. There is a little bit of duplicate code in *templates/generator.hbs* to render tge initial state. We could remove the duplication by rerendering the template upon every model change, but I had a little trouble with this. I could investgate that more thoroughly if you think it's worth it.

app/scripts/views/generator-view.js
GrantMoyer: I think we should move the simpleCopy check and createHiddenInput call into the CopyPaste.copy function then.

app/scripts/util/tip.js
antelle: I mean, not the OS tooltip, but the one KeeWeb has: https://github.com/keeweb/keeweb/blob/master/app/templates/generator.hbs#L3. When you add `title` attribute, it will render a custom tooltip. What's the need of adding this additional logic on top of it?

app/scripts/views/generator-view.js
antelle: I see, I thought that was added to render a different tooltip that will be shown _after_ you have clicked the button. Nvm, ignore my comment then.

app/scripts/views/generator-view.js
GrantMoyer: d96f458

app/scripts/models/app-settings-model.js
GrantMoyer: d96f458

app/scripts/views/generator-view.js
GrantMoyer: d96f458

app/scripts/util/tip.js
GrantMoyer: The custom tooltips are created from the title attributes of a jquery element `e` and all it's children when `Tip.createTips(e)` is called. Unfortunately, it's only called once, but the tooltip needs to be recreated each time the toggle button is pressed, since the tooltip changes back and forth from "Hide password" to "Show password". Setting the title attribute doesn't automatically regenerate the tooltip.

app/scripts/views/generator-view.js
GrantMoyer: Let me know if 4892e7a suitably addresses this.

app/scripts/views/generator-view.js
GrantMoyer: Marking this resolved because I think it's the same as https://github.com/keeweb/keeweb/pull/1223#discussion_r312105527

app/scripts/util/tip.js
GrantMoyer: I just realized that you're saying you'd prefer a non-changing tooltip that says "Hide/show password" to reduce code complexity. Is that correct?

app/scripts/views/generator-view.js
antelle: No, please check `!CopyPaste.simpleCopy` here, modifying copy-paste will break other stuff

antelle:
I've merged everything and added some fixes. The discussion already takes long and I don't want to spend your time on this.  
Please check if everything is working as expected on [beta.keeweb.info](https://beta.keeweb.info).

Thanks for the contribution! 🎁

GrantMoyer:
Looks like everything works as expected to me.

antelle:
Nice, then I'll make a release with it and other fixes soon

Issue #1190, opened 2019-06-06
[Feature Request] Copy OTP into clipboard on Autotype


dploeger:
**Is your feature request related to a problem? Please describe.**
On mfa-enabled apps, I always need to manually search for the current otp token after using autotype to automatically log into the app.

**Describe the solution you'd like**
I saw this in Kypass, which has the same problems as iOS is only supporting a username/password based login. They solved that by simply copying the OTP into the clipboard before returning the username and password.

So, when Autotype takes place and the entry has an otp, simply copy that otp into the clipboard. This should be configurable of course.

**Describe alternatives you've considered**
You could as well use the custom Autotype sequence with the OTP enabled, but this is cumbersome because many applications handle the use of OTP differently.

jivarson:
Either that or create a hotkey for "autotype TOTP". With TOTP chaning every 30th second it could be useful. Perhaps <Shift> + Enter? or maybe even better, let <Shift>+Enter (or some other combo) be up for the user to bind..

dploeger:
Will be handled by #1243

snakyjake1:
Did this feature ever get implemented?
If so, how do I get it to work?
I want to have the OTP copied to the clipboard on autotype.

Thanks.

antelle:
@snakyjake1 as you can see in comments, it was implemented as an alternative solution described in the linked issue.

snakyjake1:
I would like a simple way (or automated) to populate the OTP field on a website.  Something that auto-populates the OTP field just like it does with the username and password.  At the very least, after populating the username+password on a website, KeeWeb can copy the OTP field into the clipboard.

antelle:
We should develop a browser extension...

Issue #963, opened 2018-07-30
Feature Request: Add a keyboard shortcut to Copy OTP in background


cdsimpkins:
Just a minor inconvenience gripe.  

When using accounts that have an OTP field, I have to come back to the app to get the OTP, but I use the background copy/paste shortcuts for user/pass.  Can you add a similar shortcut for the OTP field?  

Thanks!

tbradsha:
This would be great!

tbradsha:
If this gets added, can you add a OTP shortcode to the account selection popup as well?

antelle:
The bot was misconfigured, sorry for a wrong notification.

tbradsha:
Great! Just waiting for v1.11 to come out to test it!

antelle:
@tbradsha v1.11 is released

antelle:
So now, when we have configurable shortcuts, it will be possible to enable it, however it will be disabled by default.

Issue #1243, opened 2019-08-25
Dynamic AutoType for any field


jivarson:
Instead of hijacking OP's thread in #1203, I'm creating a feature request here.

The clipboard in most OS's are insecure and not suitable for the delivery of secrets.
KeeWeb already has an excellent AutoType-implementation which is the foremost reason I use it.

Currently it's easy to autotype username and password dynamically. Other fields has to be set up beforehand.

This sort of limits the use of autotype when there's many secrets for an item.

I'd like AutoType to be more dynamic on this.

### - While in background;
In the dialogue of Global Auto-type, when the correct item is selected, press and hold Shift to browse the items Fields including custom ones. Releasing <Shift> selects and autotypes it.

### - While in foreground:
Holding <Shift> while clicking a Field of an entry autotypes it instead of copying it to clipboard.

antelle:
This makes a lot of sense, however it should be implemented as a soft of field action icons that will appear on hover (which will include copy, auto-type, show/hide password, etc...).

jivarson:
Indeed, it's better to present functionality visually rather than using modifier-keys, this would work in foreground-mode.
But I'd really like to have it accessible via autotype in background as well, without the need of a mouse.

antelle:
Sorry I read it wrong and first time I thought the suggestion was to auto-type on shift-click. So it should be:

- `background` auto-type: right, it should be a new hotkey combo for this, like shift or something else.
- `foreground` auto-type: field icons. Clicking on a field maybe doesn't do anything at all, but doesn't perform auto-type (this would be very unexpected, while copying is somewhat ok, but unexpected too, but it doesn't send the password anywhere).

jivarson:
Correct.
Key-combo is needed in background autotype, while floating icons or something would do it in foreground autotype

Coises:
Dragging and dropping the label works now on Windows.  (I seem to recall being told that it doesn't work on Mac, though.)  For me, at least, that covers the situation where I have room to get both KeeWeb and the target window on screen at the same time, and it avoids taking focus away from the target window once you've located the correct entry in KeeWeb and switched back to the target window.

I use this method all the time for entering data that doesn't fit in the "User/Password" mold.

Not that an auto-type method for non-standard fields wouldn't be good to have... I just wanted to note that there is a solution other than copy/paste that works now given the right conditions.

jivarson:
Awesome work!

antelle:
[...]

PR #1249, opened 2019-09-03
Copy OTP support in autotype. Was requested in #1190 and #963

dploeger:
Hey antelle!

It's me again implementing a feature I need 😉 
This was partly requested in #1190 and #963. It supports only inserting the OTP in autotype much like the existing features to insert only the username and password.

I have selected CTRL + Enter for this feature. Please take a look.


antelle:
Hi! Let's do it this way, the list is becoming really big and not pretty: https://github.com/keeweb/keeweb/issues/1243

Issue #1125, opened 2019-02-27
Feature Request: Add Shortcut for showing password when key is pressed


ramonreschke:
It is a bit inconvenient in KeeWeb to show a password for example to another user. The current possibility is to click on the password and go into the editing password. If I do that, I'll risk to change the password.

It would be nice to have an shortcut like pressing the `ALT` key to show the password. During the keypress the password will be shown, if the keypress ends the password will be hidden again.

joaopms:
I don't know if you're talking about an entry password or the master password, but both would be nice!

antelle:
Deployed to [beta.keeweb.info](https://beta.keeweb.info), will be available in the next release.

<img width="307" alt="Screen Shot 2019-09-26 at 23 45 40" src="https://user-images.githubusercontent.com/633557/65727391-e65cf700-e0b7-11e9-9c64-d5531c751ce9.png">

PR #1268, opened 2019-09-26
Issue 1125

dploeger:
This is a suggestion for #1125. When you hold alt and hover over a field, then after a second the value is shown. If you move the mouse out, the value is hidden again.

Let me know what you think. Especially the security implications have to be checked. I know you took some great lengths to hide the values from browser memory.

Also, I added the second delay on purpose to suppress the function if used by accident. In the current implementation, the alt key has to be pressed when the mouse reaches the field. If the hover starts and the alt key is pressed after that, it doesn't work. That maybe confusing, right?


antelle:
Hi! This will be implemented in a different way, already in progress.

PR #1297, opened 2019-10-11
remove duplicate apple-mobile-web-app-capable meta tag

coliff:
the meta tag:
`<meta name="apple-mobile-web-app-capable" content="yes" />`
was appearing twice


antelle:
Thank you!

PR #1348, opened 2019-11-13
Remove "binning" from password generation

GrantMoyer:
This will remove the "binning" from password generation, therefore increasing the entropy. Generated passwords will no longer necessarily have an equal number of characters from each selected set.


antelle:
Thanks! 🎁  merged it to develop

PR #1394, opened 2020-01-18
Add support for building KeeWeb on Windows

bradchristensen:
As the title suggests, it looks like this was once working but presumably Windows got left behind at some point 😄 

- Configured line ending settings in the Prettier config to "auto"
  - Without this, Prettier wants to convert everything to Unix line endings because it inherits that config from the `.editorconfig` - this causes linting to fail and therefore the `npm run dev` command doesn't work. The downside to this is that Prettier will no longer ensure all files use Unix line endings (it will only ensure that line endings are consistent within a file).
  - Since the repository doesn't include a `.gitattributes` file, files are checked out on Windows with CRLF line endings by default. However, any commits made on Windows will convert those files back to Unix (LF) line endings for git's internal representation of the files, when using git's default line endings config.
  - A proper fix would be to introduce a `.gitattributes` file containing `* text eol=lf` but this would require normalizing the line endings for all of the files in the repo
- Only try to load code signing config on macOS
- If loading the code signing config file fails, print a warning and carry on anyway
  - I can understand that it'd be useful to have this throw an error for non-dev builds, but I couldn't think of an easy way to detect this unless you run builds with code signing on a CI server (in which case you could check for a CI-specific environment variable, e.g. `if (process.env.CI) throw new Error(...)`)
- Fixed a regular expression in the Webpack config to support Windows path separators (backslashes)
- Fixed the `npm run electron` script to set an environment variable in a cross-platform way (using `cross-env`)

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,19 +3,24 @@
 const fs = require('fs-extra');
 const path = require('path');
 const debug = require('debug');
-const os = require('os');
 
 const webpackConfig = require('./build/webpack.config');
 const webpackConfigTest = require('./test/test.webpack.config');
 const pkg = require('./package.json');
 
+const skipCodeSigning = process.argv.some(arg => arg.startsWith('--no-sign'));
 let codeSignConfig;
-if (os.platform() === 'darwin') {
+
+if (!skipCodeSigning) {
     try {
         codeSignConfig = require('../keys/codesign');
     } catch (err) {
-        // eslint-disable-next-line no-console
-        console.warn('Code signing config missing - signing the macOS build will be skipped.');
+        throw new Error(
+            'Unable to load code signing config from ../keys/codesign.\n' +
+                'This is needed for production builds targeting macOS.\n' +
+                'For development builds, run with the `--no-sign` arg to skip code signing,\n' +
+                'e.g. `npm start -- --no-sign`'
+        );
     }
 }
 
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -3,11 +3,21 @@
 const fs = require('fs-extra');
 const path = require('path');
 const debug = require('debug');
+const os = require('os');
 
 const webpackConfig = require('./build/webpack.config');
 const webpackConfigTest = require('./test/test.webpack.config');
 const pkg = require('./package.json');
-const codeSignConfig = require('../keys/codesign');
+
+let codeSignConfig;
+if (os.platform() === 'darwin') {
+    try {
+        codeSignConfig = require('../keys/codesign');
+    } catch (err) {
+        // eslint-disable-next-line no-console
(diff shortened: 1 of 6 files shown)


Gruntfile.js
```diff
@@ -3,11 +3,21 @@
...
 
 const webpackConfig = require('./build/webpack.config');
 const webpackConfigTest = require('./test/test.webpack.config');
 const pkg = require('./package.json');
-const codeSignConfig = require('../keys/codesign');
+
+let codeSignConfig;
+if (os.platform() === 'darwin') {
+    try {
+        codeSignConfig = require('../keys/codesign');
```
antelle: Let's add a command-line option called `--no-sign` instead of this.

Gruntfile.js
```diff
@@ -3,11 +3,21 @@
...
+    } catch (err) {
```
antelle: Remove this catch: if the option is not passed, it's better to fail hard, a person building the app is not expected to monitor build logs.

Gruntfile.js
```diff
@@ -7,7 +7,22 @@ const debug = require('debug');
 const webpackConfig = require('./build/webpack.config');
 const webpackConfigTest = require('./test/test.webpack.config');
 const pkg = require('./package.json');
-const codeSignConfig = require('../keys/codesign');
+
+const skipCodeSigning = process.argv.some(arg => arg.startsWith('--no-sign'));
```
antelle: Should be
```javascript
grunt.option('no-sign')
```

bradchristensen: Whoops, sorry! I'm a Grunt newb 😅

Gruntfile.js
bradchristensen: As you've probably already noticed, I've kept the try/catch but only to be able to throw a more detailed Error with info on how to get around it if need be 🙂

Gruntfile.js
```diff
@@ -3,12 +3,13 @@
...
 const path = require('path');
 const debug = require('debug');
+const grunt = require('grunt');
 
 const webpackConfig = require('./build/webpack.config');
 const webpackConfigTest = require('./test/test.webpack.config');
 const pkg = require('./package.json');
 
-const skipCodeSigning = process.argv.some(arg => arg.startsWith('--no-sign'));
+const skipCodeSigning = grunt.option('no-sign');
```
antelle: Not like this. Remove it from this block and put it inside `module.exports = function(grunt)`, where `grunt` instance is available. You can move reading the config there as well.

Gruntfile.js
antelle: Yes noticed this, it makes sense indeed.

Gruntfile.js
bradchristensen: My bad - I should have noticed that 🤦‍♂

antelle:
Thanks! 🎁  Merged it.

Issue #1360, opened 2019-12-02
Electron or PWA?


Stephan-P:
KeeWeb Desktop is Electron based and through the KeePassHttp-Connector it can provide Autofill facilities in browsers.

Modern webbrowsers are starting to offer functionality to install a website as a PWA. 
I just used Edge Dev to install app.keeweb.info as a PWA on my W10 laptop. 
Unfortunately I was not able to install the KeePassHttp-Connector plugin in this setting and so the autofill connection was not (yet) possible.

This brings up a question of future direction:
- To continue along the Electron path
- or change direction to PWA for local installation KeeWeb?

antelle:
Hi!
KeeWeb can be actually added as PWA on iOS and Android. Installing as PWA however don't give a lot of advantage on desktop and now it has less possibilities compared to Electron.
So, for example, as you already noticed, it's not possible (and I guess, won't be possible in the nearest future) to use things like KeePassHttp-Connector or something else.
Actually I don't understand the question or proposal being suggested:
- Shall we switch to PWA on desktop? No.
- Do we need to support PWA? Yes, for example, on ChromeOS it's the only way to install an "app" (I haven't checked if it works on ChromeOS though, most likely not).

Stephan-P:
And through what method can a connection be established between Keeweb PWA and the webbrowser in order to obtain autofill functionality?

antelle:
I don’t know if it will appear in future, it’s possible. For now there’s no such method.

loginov-rocks:
Despite the initial question was rather about the desktop version, I want to highlight, that https://app.keeweb.info isn't suggested as installable either on Android, neither on macOS. May we track this issue in here?

antelle:
@loginov-rocks I didn't get it, why it's not suggested? For example, iOS shows a suggestion to install it in Sharing, which should also work on Android.

loginov-rocks:
@antelle I have never got an installation suggestion on Android One (pure 9, 10) device.

PR #1405, opened 2020-02-06
Make the app installable as PWA

loginov-rocks:
In order to be installable as PWA, the app should have at least **start_url** property in the manifest.

To address https://github.com/keeweb/keeweb/issues/1360#issuecomment-582912644

Also checked with Lighthouse audit, no PWA blocking issues found.

### DIFF
--- a/app/manifest/manifest.json
+++ b/app/manifest/manifest.json
@@ -2,6 +2,7 @@
     "name": "KeeWeb",
     "short_name": "KeeWeb",
     "description": "Free cross-platform password manager compatible with KeePass",
+    "start_url": "./",
     "display": "standalone",
     "orientation": "any",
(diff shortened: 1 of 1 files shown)


antelle:
Hi! Do you have a reference? I can’t find jt anywhere in docs, I mean, the property does exist and allows to set the url, but it’s not mandatory and there’s nothing about ‘installable’. So what does it change?

loginov-rocks:
@antelle you're right, it's hard to find anything about that in the documentation. However, I don't receive proper behavior and Chrome Dev Tools highlights the following: **Manifest start URL is not valid**.

Screen Shot 2020-02-10 at 3 29 30 PM

antelle:
Ok makes sense, thanks!

PR #1429, opened 2020-03-14
fix #1293 - copying website omits protocol 

### LINKED ISSUE #1293: Copying "Website" omits the protocol part of the url
**Describe the bug**
When clicking on "Website" in the gui, the protocol part of the url is omitted (only the host and path of the url is copied to the clipboard).

**To Reproduce**
1. Click on `+`
2. Select `Entry` to create a new entry
3. Enter a `Website`, e.g. `https://www.gmail.com`
4. Click tab or click some other field so that the website is persisted
5. Click on `Website` again to copy the url to the clipboard
6. Open a text editor and paste

**Expected behavior**
The full url, including the protocol, should be pasted, e.g. `https://www.gmail.com`

**Actual behaviour**
The protocol part is missing when pasting the url, e.g. `www.gmail.com`

**Comment**
This is not a problem if the website has a redirect from http to https (gmail does this), otherwise the result is a connection error. Since KeeWeb has no way of knowing this, I suggest that you always copy the full url.

**Environment**
KeeWeb v1.11.5 (e8e2b499, 2019-09-29)
Environment: electron v6.0.10
User-Agent: Mozilla/5.0 (Macintosh; Intel Mac OS X 10_14_6) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.11.5 Chrome/76.0.3809.146 Electron/6.0.10 Safari/537.36

harshlele:
Fixes #1293 

Also, one thing i noticed is that when the website field is unfocused it doesn't show the protocol, but it only does this for https:// urls. All other protocols like http:// or ftp:// are shown, which seems a bit weird.

### DIFF
--- a/app/scripts/views/fields/field-view.js
+++ b/app/scripts/views/fields/field-view.js
@@ -83,18 +83,30 @@ class FieldView extends View {
         let copyRes;
         if (field) {
             const value = this.value || '';
-            if (value && value.isProtected) {
-                const text = value.getText();
-                if (!text) {
+            if (value) {
+                if(value.isProtected){
+                    const text = value.getText();
+                    if (!text) {
+                        return;
+                    }
+                    if (!CopyPaste.simpleCopy) {
+                        CopyPaste.createHiddenInput(text);
+                    }
+                    copyRes = CopyPaste.copy(text);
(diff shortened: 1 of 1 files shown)


antelle:
too much duplicated code

PR #1432, opened 2020-03-19
Fix: MAINTAINER directive in Dockerfile is deprecated

canthar:
Source: https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


antelle:
🎁

PR #1472, opened 2020-04-20
add parameter to catch statement

aplgithub:


antelle:
any explanation?

aplgithub:
Appologies; the project was not compiling on my host without this modification, but I have just tried to get the error now and it seems to compile -- I am unfortunately not able to get back the environment where it was not compiling either.  I'll close this PR.

PR #1511, opened 2020-05-21
Fixed auto-type on tiling wm

b3nj5m1n:
Fix for issue with auto-type on a tiling window manager.
Grunt passes without errors, tested browser version (Not affected by changes) and desktop version (On linux & windows, macos shouldn't be any different from windows)

- The id for the window that was active at the time of pressing the hotkey gets passed to auto-type-emitter-linux.js where it is passed as an argument to xdotool, which causes the tool to send the keystrokes directly to the desired window, regardless of the active window.
- The window id only gets passed to the emitter if the platform is linux, the code for other platforms was not modified and should therefore work just as before.
- In index.js I commented out an if statement testing whether or not the active window id before selecting a login in keeweb was the same afterwards. Maybe this statement should be put back in and only throw an error if the tool for simulating keystrokes does not support sending those to a specific window.

### DIFF
--- a/app/scripts/auto-type/auto-type-emitter-factory.js
+++ b/app/scripts/auto-type/auto-type-emitter-factory.js
@@ -1,14 +1,11 @@
 import { Launcher } from 'comp/launcher';
 
 const AutoTypeEmitterFactory = {
-    create(callback, windowID) {
+    create(callback, windowId) {
         if (Launcher && Launcher.autoTypeSupported) {
             const { AutoTypeEmitter } = require('./emitter/auto-type-emitter-' +
                 Launcher.platform());
-            if (Launcher.platform() === 'linux') {
-                return new AutoTypeEmitter(callback, windowID);
-            }
-            return new AutoTypeEmitter(callback);
+            return new AutoTypeEmitter(callback, windowId);
         }
         return null;
     }
--- a/app/scripts/auto-type/auto-type-emitter-factory.js
+++ b/app/scripts/auto-type/auto-type-emitter-factory.js
@@ -1,10 +1,13 @@
 import { Launcher } from 'comp/launcher';
 
 const AutoTypeEmitterFactory = {
-    create(callback) {
+    create(callback, windowID) {
         if (Launcher && Launcher.autoTypeSupported) {
             const { AutoTypeEmitter } = require('./emitter/auto-type-emitter-' +
                 Launcher.platform());
+            if (Launcher.platform() === 'linux') {
+                return new AutoTypeEmitter(callback, windowID);
+            }
             return new AutoTypeEmitter(callback);
         }
         return null;
--- a/app/scripts/auto-type/auto-type-parser.js
+++ b/app/scripts/auto-type/auto-type-parser.js
@@ -1,7 +1,8 @@
 import { AutoTypeRunner } from 'auto-type/auto-type-runner';
 
-const AutoTypeParser = function(sequence) {
+const AutoTypeParser = function(sequence, windowID) {
     this.sequence = sequence;
+    this.windowID = windowID;
     this.ix = 0;
     this.states = [];
 };
@@ -42,7 +43,7 @@ AutoTypeParser.prototype.parse = function() {
     if (this.states.length !== 1) {
         throw 'Groups count mismatch';
     }
-    return new AutoTypeRunner(this.state().ops);
+    return new AutoTypeRunner(this.state().ops, this.windowID);
 };
 
 AutoTypeParser.prototype.pushState = function() {
--- a/app/scripts/auto-type/auto-type-parser.js
+++ b/app/scripts/auto-type/auto-type-parser.js
@@ -1,8 +1,7 @@
 import { AutoTypeRunner } from 'auto-type/auto-type-runner';
 
-const AutoTypeParser = function(sequence, windowID) {
+const AutoTypeParser = function(sequence) {
     this.sequence = sequence;
-    this.windowID = windowID;
     this.ix = 0;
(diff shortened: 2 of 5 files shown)


app/scripts/auto-type/auto-type-emitter-factory.js
```diff
@@ -1,10 +1,13 @@
 import { Launcher } from 'comp/launcher';
 
 const AutoTypeEmitterFactory = {
-    create(callback) {
+    create(callback, windowID) {
         if (Launcher && Launcher.autoTypeSupported) {
             const { AutoTypeEmitter } = require('./emitter/auto-type-emitter-' +
                 Launcher.platform());
+            if (Launcher.platform() === 'linux') {
```
antelle, line 8: This check is not needed: if `windowId` is provided, let's use it. Other systems may use it too in future.

app/scripts/auto-type/auto-type-emitter-factory.js
```diff
@@ -1,10 +1,13 @@
 import { Launcher } from 'comp/launcher';
 
 const AutoTypeEmitterFactory = {
-    create(callback) {
+    create(callback, windowID) {
```
antelle, line 4: Please rename to `windowId` to follow the naming convention

app/scripts/auto-type/auto-type-parser.js
```diff
@@ -1,7 +1,8 @@
 import { AutoTypeRunner } from 'auto-type/auto-type-runner';
 
-const AutoTypeParser = function(sequence) {
+const AutoTypeParser = function(sequence, windowID) {
```
antelle, line 3: The parser doesn't need this data, let's remove it from here and pass to the runner in its `run` method

app/scripts/auto-type/index.js
```diff
@@ -185,20 +186,21 @@ const AutoType = {
...
                 return callback(false);
             }
-            if (activeWindowInfo.id !== windowInfo.id) {
-                logger.info(
-                    `Active window doesn't match: ID is different. ` +
-                        `Expected ${windowInfo.id}, got ${activeWindowInfo.id}`
-                );
-                return callback(false, activeWindowInfo);
-            }
+            // if (activeWindowInfo.id !== windowInfo.id) {
```
antelle, line 189: This should be disabled only if `windowId` is present

b3nj5m1n:
I think I addressed all of your requested changes, as for

> This should be disabled only if windowId is present

The variable will be present on other operating systems as well, so for now I've added a condition to check if the os is linux, since all versions of linux should use xdotool and therefore support sending keystrokes to a specific window.

I've only had time to check on linux, but it seems to work fine and I don't see why it would be different on another os.
If there is anything else you would like me to change, or if I've missed something, let me know.

app/scripts/auto-type/index.js
```diff
@@ -185,7 +185,8 @@ const AutoType = {
                 logger.debug('Error during active window check, something is wrong', err);
                 return callback(false);
             }
-            if (activeWindowInfo.id !== windowInfo.id) {
+            this.windowId = windowInfo.id;
```
antelle, line 188: It shouldn't be assigned to `this`, especially in `activeWindowMatches` method, this is an unexpected side-effect. You have access to `windowInfo` in the caller, it should pass it to `runAndHandleResult` directly.

b3nj5m1n: Sorry for my bad knowledge of js, I've made another commit to fix this, I hope it's okay now.

app/scripts/auto-type/index.js
```diff
@@ -58,8 +58,8 @@ const AutoType = {
         }
     },
 
-    runAndHandleResult(result) {
-        this.run(result, err => {
+    runAndHandleResult(result, windowId) {
```
antelle, line 61: This function is called without `windowId` in one place in this file, looks like it will fail there.

b3nj5m1n: If I'm not mistaken (currently on my phone so can't check) that is the call for when a matching entry was found automatically, therefore the window to type to is already the active window and the id doesn't necessarily have to be passed.
I think I still tried to put it in there but I didn't have access to windowInfo, but maybe I'm just misremembering something.
If the Id really has to be passed there, I'll take a look at it in a few hours or something.

antelle: Well, you can make it optional, but if so, you need to handle it [here](https://github.com/keeweb/keeweb/pull/1511/files#diff-e3232c8bbd07c6d2be79351d60d87a21R80), otherwise it will pass empty `--window` arg to `xdotool` and I guess, `xdotool` won't like this.

b3nj5m1n: Actually I just tested it and if you pass undefined as the window argument it just uses the active window, so I don't think it would really be necessary, if you still want it I can implement it.

antelle: It's better not to pass `undefined`, this can have unexpected effects and it _looks like_ that the parameter is required. If it's optional, let's handle it in the key-emitter and explicitly remove it from the command line, instead of inserting `undefined` there.

b3nj5m1n: It should be fixed now.

antelle: Awesome, thank you!

antelle:
Thank you! 🎁 
Merged it to `develop` in b715507dd0ae5485c7629cdf57e943e84e67cb9d, it will be included in the next version.
Please branch from `develop` next time, then GitHub would mark the pull request as merged, now I'll have to close it.

Issue #1564, opened 2020-07-14
Linux Auto-type no longer works in 1.15.x series


opennomad:
Auto-type (ALT-SHIFT-T) no longer works in linux. I tested with 1.15.5, 1.15.4, and 1.15.0. None of them appear to run auto-type. When I hit the hot-key it focuses KeeWeb and I can select the password. When I hit enter it goes back to where I triggered the hot-key, but then nothing happens.

1.14.3 does not have that issue.

KeeWeb v1.15.5 (62ea631, 2020-06-13)
Environment: electron v9.0.4
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.15.5 Chrome/83.0.4103.104 Electron/9.0.4 Safari/537.36

antelle:
Hi! Is there anything interesting in logs?

opennomad:
here's what I see:

```
[DEBUG] 2020-07-14T18:33:30.546Z [auto-type] Hide window
[DEBUG] 2020-07-14T18:33:30.653Z [auto-type] Getting window info
[INFO ] 2020-07-14T18:33:30.670Z [launcher] spawn xdotool: 0, 18ms
us-east-1.signin.aws.amazon.com - Amazon Web Services Sign-In - Mozilla Firefox
85983235
[DEBUG] 2020-07-14T18:33:30.671Z [auto-type] Window info 85983235 us-east-1.signin.aws.amazon.com - Amazon Web Services Sign-In - Mozilla Firefox null
[INFO ] 2020-07-14T18:33:30.671Z [auto-type] Active window matches
[DEBUG] 2020-07-14T18:33:30.671Z [auto-type] Start {USERNAME}{TAB}{PASSWORD}{ENTER}
[DEBUG] 2020-07-14T18:33:30.672Z [auto-type] Parsed [op:USERNAME,op:TAB,op:PASSWORD,op:ENTER]
[DEBUG] 2020-07-14T18:33:30.673Z [auto-type] Resolved [********,key:tab,[*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*,*],key:enter]
[INFO ] 2020-07-14T18:33:30.880Z [launcher] spawn xdotool: 0, 207ms
[INFO ] 2020-07-14T18:33:30.927Z [launcher] spawn xdotool: 0, 46ms
[...]

PR #1594, opened 2020-09-08
use windowactivate via xdotool

opennomad:
I'm making this PR against `master` since that is where I'm having issues. This PR only changes the use of `xdotool` in the linux auto-type emitter, but using `windowactivate` to bring the window up. With these changes auto-type on LInux succeeds again both with `alt-shift-t` and `ctrl-t`.

I was unable to run the linting correctly and so I uncommented that to allow me to implement this hotfix.

This is a single file change only and impacts only the linux platform.

This resolves issue #1564

### DIFF
--- a/app/scripts/auto-type/emitter/auto-type-emitter-linux.js
+++ b/app/scripts/auto-type/emitter/auto-type-emitter-linux.js
@@ -61,13 +61,11 @@ const ModMap = {
 
 const AutoTypeEmitter = function (callback, windowId) {
     this.callback = callback;
-    if (typeof windowId !== 'undefined' && windowId) {
-        this.windowParameter = '--window ' + windowId + ' ';
-    } else {
-        this.windowParameter = '';
-    }
     this.mod = {};
     this.pendingScript = [];
+    if (typeof windowId !== 'undefined' && windowId) {
+        this.pendingScript.push('windowactivate --sync ' + windowId);
+    }
 };
 
 AutoTypeEmitter.prototype.setMod = function (mod, enabled) {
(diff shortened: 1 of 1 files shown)


app/scripts/auto-type/emitter/auto-type-emitter-linux.js
```diff
@@ -61,13 +61,11 @@ const ModMap = {
 
 const AutoTypeEmitter = function (callback, windowId) {
     this.callback = callback;
-    if (typeof windowId !== 'undefined' && windowId) {
-        this.windowParameter = '--window ' + windowId + ' ';
```
antelle, line 65: I'm thinking, does it break if we leave `windowParameter` as well? For example, if a window failed to activate for some reason, or if another window was opened during auto-type, keystrokes may be sent to another application. In theory it should work with both, but if it doesn't, let's remove it for sure.

antelle:
Thanks a lot for looking into it! 🎁 
Left a small question about the window parameter, apart from this it looks good to me.

app/scripts/auto-type/emitter/auto-type-emitter-linux.js
opennomad: I've been running the code for the past 2 days and it's been solid for me.

In all my tests (via app and xdotool CLI), I just can't get `xdotool` to do the right when specifying the window. For example, in a terminal, I can run `xdotool type --window <terminals window_id> "hello"` (targeting the window I'm already in) and nothing happens. It only works if I remove that parameter. That suggests to me that leaving it off is the only way it works.

The key appears to be in the "SENDEVENT NOTES" section of the xdotool manpage:

> If you are trying to send key input to a specific window, and it does not
> appear to be working, then it's likely your application is ignoring the events
> xdotool is generating.  This is fairly common.
> 
> Sending keystrokes to a specific window uses a different API than simply typing
> to the active window. If you specify 'xdotool type --window 12345 hello'
> xdotool will generate key events and send them directly to window 12345.
> However, X11 servers will set a special flag on all events generated in this
> way (see XEvent.xany.send_event in X11's manual). Many programs observe this
> flag and reject these events.
> 
> It is important to note that for key and mouse events, we only use XSendEvent
> when a specific window is targeted. Otherwise, we use XTEST.
> 
> Some programs can be configured to accept events even if they are generated by
> xdotool. Seek the documentation of your application for help.
> 
> Specific application notes (from the author's testing): * Firefox 3 seems to
> ignore all input when it does not have focus.  * xterm can be configured while
> running with ctrl+leftclick, 'Allow SendEvents' * gnome-terminal appears to
> accept generated input by default.

That suggest that the use of the `--window` parameter using the XsendEvent will not work for all apps, while the XTEST does appear to work universally. To me that suggests that leaving the option out is the better choice.

Using both via config may also be doable, but might just lead to user confusion and frustration.

antelle: Makes sense, thanks a lot for digging that far! I don't think it needs to be configurable since it seems to be broken for you.

antelle:
Merged into `release-1.15` in 31cc2fbdd4fc5a2df1049190d849e21b6924654a, I'll make a release with it this week.

PR #1631, opened 2020-11-26
DateFormat is based on specified Locale

### LINKED ISSUE #1606: Change Date format ..
An option to Change Date format .. like 2020-09-22 OR 2020/09/22 Thir .. OR maybe called [ISO date Format ](https://en.wikipedia.org/wiki/ISO_8601)

MrSnix:
**Changes**:
1. Introduced `Intl.DateTimeFormat` in `date-format.js`
2. Rewritten `dtStr`, `dStr`and `dtStrFs`
3. Removed hard-coded locales `months`, `weekdays` and `shortWeekdays` from `base.json`
4. Replaced `months`, `weekdays` and `shortWeekdays` values from `base.json` with the following functions` getMonthsForLocale()`, `getWeekDaysForLocale()` and `getWeekdaysShortForLocale()`. Now are dynamically generated using locale.

**Any feedback?** 
Let me know if i messed up something, i'm just new to codebase.

Closes #1606


app/scripts/util/formatting/date-format.js
```diff
@@ -1,19 +1,46 @@
-import { StringFormat } from 'util/formatting/string-format';
-import { Locale } from 'util/locale';
+import { SettingsManager } from 'comp/settings/settings-manager';
```
antelle, line 1: Modules in `utils` should not use stateful components from comp, perhaps we need to move this file since it's becoming stateful.

app/scripts/util/formatting/date-format.js
```diff
@@ -22,27 +49,19 @@ const DateFormat = {
...
-                  '-' +
-                  StringFormat.pad(dt.getDate(), 2) +
-                  'T' +
-                  StringFormat.pad(dt.getHours(), 2) +
-                  '-' +
-                  StringFormat.pad(dt.getMinutes(), 2) +
-                  '-' +
-                  StringFormat.pad(dt.getSeconds(), 2)
-            : '';
+        return dt ? new Date().toISOString().replaceAll(':', '-').slice(0, 19) : '';
```
antelle, line 64: This will always print the current date if I'm not mistaken. There are tests for this module, they also need an update I guess.

antelle (changes requested):
Thanks! 🎁  added some comments

app/scripts/util/formatting/date-format.js
antelle: If a method accepts date as a parameter, it should print exactly this date, not the current date. I mean, look, you do `new Date()` if `dt` is passed, that's the mistake, you meant `dt` here right? Something like `return dt ? dt.toString... : ''`

MrSnix: Yes @antelle, i have **already fixed that** in the next commit. I didn't notice it before :D

test/src/util/formatting/date-format.js
```diff
@@ -1,15 +1,15 @@
 import { expect } from 'chai';
-import { DateFormat } from 'util/formatting/date-format';
+import { DateFormat } from 'comp/util/date-format';
 
 describe('DateFormat', () => {
-    const dt = new Date(2020, 0, 2, 3, 4, 5, 6);
+    const dt = new Date.UTC(2020, 0, 2, 3, 4, 5, 6);
```
antelle, line 5: Two things here:
1. Does this work? Date.UTC returns a number, it shouldn't be possible to use it as a constructor.
2. Printing dates in UTC is not correct, we want to display dates in the local time, not in UTC. This test must print a different value in this case. But better not to change the test.

test/src/util/formatting/date-format.js
```diff
@@ -1,15 +1,15 @@
...
 
     it('should format date', () => {
-        expect(DateFormat.dStr(dt)).to.eql('2 Jan 2020');
+        expect(DateFormat.dStr(dt)).to.eql('Jan 2, 2020');
     });
 
     it('should format date and time', () => {
-        expect(DateFormat.dtStr(dt)).to.eql('2 Jan 2020 03:04:05');
+        expect(DateFormat.dtStr(dt)).to.eql('Jan 2, 2020, 03:04:05 AM');
```
antelle, line 12: This test will fail if system locale is not `en-US`, it should set the locale now.

antelle: But I see, en-us is the default, so it should work... sorry

test/src/util/formatting/date-format.js
MrSnix: 1)Yes, it works. I try to remember running tests before committing. I did run `grunt test`.
image
It stills work because there is a conditional if which transform all the numbers in dates.
image
2) If i do not put date into UTC it will fail due to timezone because the function uses specifically the `activeLocale`.
If we want to make it stateful, otherwise i need to change the function signature adding the requested locale. Anyway, in these tests we are trying to see if date is printed as expected, in that case we are taking `en` as Locale (the default one) and checking if it returns as should be. Isn't it?

test/src/util/formatting/date-format.js
MrSnix: No, it doesn't fail. `en-US` is the default Locale provided by the application. I'm running an it-IT locale, still the test doesn't fail.

test/src/util/formatting/date-format.js
antelle: Ok, thanks! I'll take a look before merging, this is very strange, but I'll figure it out then.

antelle: I see, tests are broken after webpack upgrade in `develop`, only top-level tests are run and it doesn't go inside. Ok, I'll take it from here.

MrSnix: Uhmm....strange... it looks like this test is running.

antelle: It's running, but it fails exactly where I pointed out: 
```
Uncaught TypeError: Date.UTC is not a constructor
    at Suite.<anonymous> (date-format.js:5)
```
anyway, I'll fix it and look into your test expectation, this is a framework issue, tests must not fail because of one test that throws.

app/scripts/comp/util/date-format.js
```diff
@@ -0,0 +1,68 @@
...
+        }
+        return dt
+            ? new Intl.DateTimeFormat(SettingsManager.activeLocale, {
+                  year: 'numeric',
+                  month: 'short',
+                  day: 'numeric',
+                  hour: '2-digit',
+                  minute: '2-digit',
+                  second: '2-digit'
+              }).format(dt)
```
antelle, line 43: I've fixed tests, but this is a bit not correct in `en-US`. It returns `03 AM` and `13` without PM. You can fix this if you want, or I'll do it later with other fixes, I don't have time to do it today already.

app/scripts/comp/util/date-format.js
```diff
@@ -0,0 +1,68 @@
...
+                  day: 'numeric'
+              }).format(dt)
+            : '';
+    },
+
+    dtStrFs(dt) {
+        if (typeof dt === 'number') {
+            dt = new Date(dt);
+        }
+        return dt ? dt.toISOString().replaceAll(':', '-').slice(0, 19) : '';
```
antelle, line 64: I will revert this change, this prints date in UTC, not sure how it's better compared to the previous implementation.

app/scripts/comp/util/date-format.js
antelle: May I ask, why did you choose this way of specifying format, and not `timeStyle: "medium"`? I'm trying to understand the issue that you were trying to solve.

MrSnix: The previous code used to pad values in order to have always 2 digit, so i wanted to make it works as expected (always padded);
`timeStyle: "medium"`allows value like 4:00 AM instead of 04:00 AM, that's why i used this specific '2-digit'.

antelle: I see, thanks! It should be padded according to locale rules, so I’ll replace it with short or medium then.

antelle:
Merged it, thank you! 🎁  You can test it on [beta.keeweb.info](https://beta.keeweb.info), but don't be scared by design, it's awful because I replaced icons, and the new theme right now is in works, so it looks very much broken now.

MrSnix:
I hope you had a good time reviewing my code and i'm sorry if i missed some points. 
I'm pretty new to contributing. 

Thanks again

PR #1642, opened 2020-11-30
Clean up apt local cache in Dockerfile

PeterDaveHello:
Reference:

- docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

> For other items (files, directories) that do not require ADD’s tar auto-extraction capability, you should always use COPY.


PeterDaveHello:
Not sure why the title is the old one from #1638 😱  ...

antelle:
No worries, I got what it was

PR #1690, opened 2021-01-22
redirect with original port in nginx config

namtab00:
Using the docker image latest.
Keeweb config is enforced via env var **KEEWEB_CONFIG_URL** to "config.json"

With a custom SSL port (i.e. 8443) does not correctly redirect 80 > 8443.

Nginx conf vars:
- `$host only` > redirects with original host name only.
- `$http_host` > redirects with original host name + original port.

Tested and working.

If merging, please ping back when a new docker image is published, as I need to remove my custom nginx conf volume mount.

### DIFF
--- a/package/docker/keeweb.conf
+++ b/package/docker/keeweb.conf
@@ -31,5 +31,5 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
     server_name _;
-    return 301 https://$host$request_uri;
+    return 301 https://$http_host$request_uri;
 }


package/docker/keeweb.conf
```diff
@@ -31,5 +31,5 @@ server {
     listen 80 default_server;
     listen [::]:80 default_server;
     server_name _;
-    return 301 https://$host$request_uri;
+    return 301 https://$http_host$request_uri;
```
antelle, line 34: I don't understand a bit, won't this be always set to 80 because the server is listening on port 80 only, as stated two lines above.

How does this work in your setup, where does the SSL port come from here?

namtab00:
Sorry I'm currently on mobile so I can't provide screenshots, I'll try to clear things up by detailing how I've used the change in the pull request:

- Keeweb is loaded from : https://my internal.trusteddomain.com:8433
- do note the non standard port; I need to use it instead of 443
- without my change the `config.json` file fails to load, and in the browser's console there is a logged CORS error: 
`no allowed origin... for https://my internal.trusteddomain.com/config.json...`
- do note the absent port
- this means that the request for config.json file (kw-config meta tag) gets loaded on 80 which then gets redirected by nginx to 443, but fails because CORS still applies between `https://my.domain.tld:442` and 
`https://my.domain.tld:<nonstandardport>`.

With my change and the custom nginx config mounted at `/etc/nginx/conf.d/keeweb.config`, everything works correctly.

antelle:
> Keeweb is loaded from : https://myinternal.trusteddomain.com:8433
In this case, why do you get to this host?
```
listen 80 default_server;
```

It doesn't accept https connections on port 8443.

antelle:
How is the config location specified? Let's look into this, maybe this is the reason it tries a wrong host?

namtab00:
That's what I was wondering, why am I hitting 80?

Maybe the way in which the request for the  config.json file is built (after the value is read from the `kw-config` meta tag) does not account for the current custom port, and falls back to 80.

antelle:
What is the contents of your `kw-config`?

namtab00:
The precise string "config.json".

This should translate as a relative path, so `https://host:8433/config.json`, but instead is loaded from `http://host/config.json`, note the proto switch...

antelle:
Wow, how does this happen? Could you please make a screenshot of the Network tab in your browser devtools? Maybe it redirects there?

antelle:
I just checked, the config location is passed as-is to XMLHttpRequest, it doesn't have any port and protocol removal logic, so it should work. I assume, it's redirected somewhere?

namtab00:
I've tried to repro the issue I had, but I'm not seeing it anymore...

Yeah, it was probably me and the 01:00 AM context.. Reject this pull request, if I hit the issue again I'll more carefully gather diag info.

Thank you and sorry for the time loss.

PS: for anyone else reading this in the future, customizing the nginx.conf in the keeweb container is **not** necessary if _kw-config_ is set to `config.json` via KEEWEB_CONFIG_URL env var and you're using a custom port.

antelle:
Ok, good that it's fixed!

Issue #1718, opened 2021-02-20
Option to minimize KeeWeb on copy


enter-is:
Original client has a nice feature where you can auto-hide the window on copy, makes it easy to input passwords using hotkeys, but not using auto-type.

I would be happy to provide a pull request for this.

PR #1721, opened 2021-02-22
Option to minimize KeeWeb on copy

enter-is:
Option to minimize KeeWeb on copy

https://github.com/keeweb/keeweb/issues/1718

### DIFF
--- a/app/scripts/framework/views/copyable.js
+++ b/app/scripts/framework/views/copyable.js
@@ -19,6 +19,11 @@ const Copyable = {
         const msg = clipboardTime
             ? Locale.detFieldCopiedTime.replace('{}', clipboardTime)
             : Locale.detFieldCopied;
+        if (AppSettingsModel.minimizeOnCopy) {
+            setTimeout(() => {
+                Events.emit('minimize-app');
+            }, 0);
+        }
         let tip;
         if (!this.isHidden()) {
             tip = Tip.createTip(fieldLabel[0], {
--- a/app/scripts/framework/views/copyable.js
+++ b/app/scripts/framework/views/copyable.js
@@ -19,11 +19,6 @@ const Copyable = {
         const msg = clipboardTime
             ? Locale.detFieldCopiedTime.replace('{}', clipboardTime)
             : Locale.detFieldCopied;
-        if (AppSettingsModel.minimizeOnCopy) {
-            setTimeout(() => {
-                Events.emit('minimize-app');
-            }, 0);
-        }
         let tip;
         if (!this.isHidden()) {
             tip = Tip.createTip(fieldLabel[0], {
--- a/app/scripts/locales/base.json
+++ b/app/scripts/locales/base.json
@@ -443,6 +443,7 @@
     "setGenClearSeconds": "In {} seconds",
     "setGenClearMinute": "In a minute",
     "setGenMinInstead": "Minimize the app instead of close",
+    "setGenMinOnCopy": "Minimize on copy",
     "setGenLock": "Auto lock",
     "setGenLockMinimize": "When the app is minimized",
     "setGenLockCopy": "On password copy",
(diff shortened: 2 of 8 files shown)


app/scripts/framework/views/copyable.js
```diff
@@ -19,6 +19,11 @@ const Copyable = {
         const msg = clipboardTime
             ? Locale.detFieldCopiedTime.replace('{}', clipboardTime)
             : Locale.detFieldCopied;
+        if (AppSettingsModel.minimizeOnCopy) {
```
antelle, line 22: This utility is used for different purposes. It shouldn't minimize here.

app/scripts/locales/base.json
```diff
@@ -443,6 +443,7 @@
     "setGenClearSeconds": "In {} seconds",
     "setGenClearMinute": "In a minute",
     "setGenMinInstead": "Minimize the app instead of close",
+    "setGenMinOnCopy": "Minimize on copy",
```
antelle, line 446: Let's be more elaborate and mention that it's only password copy

app/scripts/views/app-view.js
```diff
@@ -603,6 +604,10 @@ class AppView extends View {
         }
     }
 
+    minimizeApp() {
+        Launcher.minimizeApp();
```
antelle, line 608: Just call it as-is, there's no need for an event

desktop/main.js
```diff
@@ -174,6 +174,10 @@ main.restartAndUpdate = function (updateFilePath) {
 };
 main.minimizeApp = function (menuItemLabels) {
     let imagePath;
+    // a workaround to correctly restore focus on windows platform
+    if (process.platform === 'win32') {
```
antelle, line 178: What exactly is broken?

antelle (changes requested):
Hi! Thanks 🎁  added a couple of comments above

app/scripts/locales/base.json
enter-is: It minimizes on all copy though. use case: open app/site requiring password, hotkey to open keeweb, find record, copy user name, paste, tab, hotkey to open keeweb again, copy password, paste, enter.

antelle: This is not how it should behave, please change it.

desktop/main.js
enter-is: on windows focus is not returned to previous field unless this workaround is added, e.g. if you are in a field on some page, open keeweb, copy (which does minimize), the focus is not returned to the field where you were before

app/scripts/locales/base.json
antelle: Sorry I read it wrong. I mean, it should minimize on copying FIELDS, i.e. what you write, exactly. But not when you copy something in a text field.

antelle: For example, if you copy password in a generator, it shouldn't minimize.

desktop/main.js
antelle: Please add this knowledge to the comment 😉

app/scripts/locales/base.json
enter-is: ah, got it, makes sense, thanks, will make the change

app/scripts/framework/views/copyable.js
enter-is: could you please point to a better place? seemed like a good place to have this code without duplication

antelle: Since it's supposed to copy after minimize only on the detail view, it can be for example, a method in details-view, which will call `fieldCopied` and then minimize the app.

enter-is: sounds good, only concern I have is history will not get this functionality, if this is not needed, I will make the change

app/scripts/locales/base.json
antelle: I think that's expected, it should copy from details, right?

enter-is: I guess this comment was to the previous thread, yes, it all makes sense, thank you

antelle (approved):
Thank you! 💯

Issue #1737, opened 2021-03-10
Simple, minimal server-side support


Coises:
KeeWeb can be used with a few different proprietary remote storage schemes, but the only non-proprietary standard is WebDAV.  I think the reason is that WebDAV appears to be the only non-proprietary standard that allows client-side ECMAScript in a browser to update a file on a server without application-specific server-side support.

Alas, support for the WebDAV standard is not always so standard.  Support on shared hosts is spotty, and if there is a simple way to add it, I couldn't find it.  Even where it is supported, other confusions (CORS seems to be a common one) cause failures.  I've gotten it to work by installing NextCloud, which feels like swatting a fly with a sledgehammer.  (At last check, #492 still applies.  This is probably a NextCloud absurdity, not a KeeWeb problem... but it demonstrates how completely flaky support for the WebDAV "standard" really is.)

Would it be possible to create a _simple_ server-side service specifically to support KeeWeb that could be installed easily on any server or shared hosting account that supports PHP?

I'm thinking that some clever use of challenge-response should make it possible to use the same passphrase that encrypts the KeePass file to protect it from access/vandalism as well, so one wouldn't have to remember separate login credentials to access the file.  For example, when a file is to be saved, in addition to the file, KeeWeb would send (over https) a randomly-generated string and the result of encrypting that same string with the passphrase.  The server-side service would store both of these in a file of the same name with a different extension.  When the KeePass file is next requested, the server would send the encrypted string and the client would respond with the unencrypted string, proving that the user supplied the correct passphrase.  Creating a new KeePass file (other than the first one) would require having already successfully opened one file.  No user name would be necessary; the file name itself would serve that role.

I don't know how broad the use for this would be, or how difficult it would be to build such a thing.  It's a shame SFTP isn't viable, but as I understand it, that won't work from a browser.

antelle:
Hi!

I'm not sure if there's demand for a simple web server, but I see why it may make sense. I'd like to get some feedback on what is desired and who would prefer setting it up this way. It should probably support more languages rather than PHP, for example, it can be also in node.js, which is quite popular now.

One way it can be implemented without a lot of effort, is using WebDAV protocol, but pre-configured, so that we have CORS headers and so on. But maybe not, depending on how people would like to authenticate there and what features we want to support, simple REST API may be more appropriate. It's better to authenticate via OAuth rather than how it's done in most of WebDAV setups.

[...]

PR #1769, opened 2021-03-29
Added docker-webdav an one possible solution for #1737

MichaelLeeHobbs:

Please follow these steps:

1. work on the right branch: it must be `develop` for new features and `master` for hotfixes
2. make sure the feature you add is not listed in [banned features](https://github.com/keeweb/keeweb/wiki/Unsupported-Features)
3. check compatibility with [the project roadmap](https://github.com/keeweb/keeweb/wiki/TODO) and key features: you should not break anything
4. if your feature is introducing a lot of new UI or changing any UX or design concept, please open an issue to discuss it first
5. please respect existing style in code, styles, and markup (hint: install and enable [editorconfig](http://editorconfig.org/) in your editor)
6. don't add any dependencies
7. test your code: it must work in all browsers, Windows, OSX, and Linux
8. if your code is platform-dependent, please add corresponding switches
9. respect each platform: don't create a behaviour that breaks interface guidelines or user expectations 
10. run `grunt` and make sure it's passing 
11. did not run grunt, does not apply

---
1. Added `package/docker-webdav`
2. Run build steps to generate dist folder
3. Run `buildDockerWebDav.sh` from `package/docker-webdav` this creates `tmp/build/dockerWebDav` and generates local docker image `keeweb:dev`
4. From the `tmp/build/dockerWebDav` folder can run `docker-compose up -d` to demo on `http://localhost:8099` and `https://localhost:44399` with the webdav at `https://localhost:44399/store/` with a copy of demo at `https://localhost:44399/store/Demo.kdbx`

Demo
WebDAV
URL: `https://localhost:44399/store/Demo.kdbx`
User: `admin`
Password: `changeme`

Getting the permissions just right on the demo took way too long. That said it does work. I don't understand your build system so I expect you will want to adjust the build. I only provided the `buildDockerWebDav.sh` as a guide for doing just that. I did this as token of good will as I expect we will drop consideration of keeweb this week.

### DIFF
--- a/package/docker-webdav/Dockerfile
+++ b/package/docker-webdav/Dockerfile
@@ -1,0 +1,53 @@
+# KeeWeb official docker container
+# https://keeweb.info
+# (C) Antelle 2019, MIT license https://github.com/keeweb/keeweb
+# Based on nginx-ssl-secure https://github.com/MarvAmBass/docker-nginx-ssl-secure/
+
+# Building locally:
+# docker build -t keeweb .
+# docker run --name keeweb -d -p 443:443 -p 80:80 -e 'DH_SIZE=512' -v $EXT_DIR:/etc/nginx/external/ keeweb
+
+# Using pre-built image from dockerhub:
+# If you have SSL certs, put your dh.pem, cert.pem, key.pem to /etc/nginx/external/ and run with:
+# docker run --name keeweb -d -p 443:443 -p 80:80 -v $EXT_DIR:/etc/nginx/external/ antelle/keeweb
+# Or, to generate self-signed cert, run:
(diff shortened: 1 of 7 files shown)


antelle:
Hi! I think an image with keeweb and webdav already exists on dockerhub.

As for support in the official image, I think it makes sense to inherit the main one, or reuse the code, because I already see that some settings are out of date (for example, SSL protocols) and maintaining them in all docker images we that have (if we have more) would be hard.

Not sure if you're still willing to do it, based on the description you provided, I assume no, so closing this. And I'll look into providing webdav support and reusing code, because it seems to be a common thing that people do.

Issue #874, opened 2018-02-23
support keepassxc browser


Akruidenberg:
It would be nice to support Keepassxc browser into the future.  Keepasshttp is less secure. an altrnative would be: https://github.com/smorks/keepassnatmsg

antelle:
Or keepass-rpc, as an alternative.

jeythekey:
Hi @antelle is there already any progress on that issue? [keepasshttp-connector](https://github.com/smorks/keepasshttp-connector) is deprecated and unchanged since almost one year.

If not, just an idea: A workaround could be to let keepassxc run in parallel and point it to the local/cached kdbx-file of keeweb. I tried it out and it works in principle, but external changes to the kdbx-file are apparently not recognized as such and hence not synced to the (webDAV) server by keeweb (they are, however, _displayed_ in keeweb immediately, but lost once I reload the file). I even cannot trigger a sync manually (with the menu button or by hitting ctrl+s).

So, would that be an easy fix to make keeweb aware of _any_ differences between the local and the remote file in order to trigger a syncing operation at the next occasion? That would be useful also in other scenarios, e.g. when using keepassxc for some features (such as removing fields) not (yet) implemented in keeweb.

Thanks for the great work by the way!

antelle:
@jeythekey the described workaround won't work for the majority of users. We should better make the plugin support native messaging. For those who would like to contribute, the plugin is here: https://github.com/keeweb/keeweb-plugins/blob/master/docs/plugins/keewebhttp/plugin.js

jeythekey:
Sure, that would be the ideal. Does your reply imply that the proposed syncing-fix is more involving? Because it would be nice also for other reasons. Thanks anyway for the quick reply!

antelle:
The proposed sync doesn't seem to be real. What we consider sync, is when the file is changed by you (or someone else) on the remote storage, then we load it, sync changes, and upload back when you make other changes. The case when you do something on the same computer with another app (which doesn't want to sync it itself for some reason) while the file is still open in keeweb, and want to sync it immediately without changes, is not the real sync.

jeythekey:
I understand, thanks. It was just unexpected, because the changes to the file are immediately visible in keeweb, so the local file known to keeweb _is_ different from the remote one, but keeweb doesn't know about that difference. Anyway, I understand this is probably something better not to loose time on, sorry for the noise.

amamabel:
+1 for a Keepassxc browser support :-)

[...]

PR #1777, opened 2021-04-03
[RFC] Chrome extension for form fill

jiggak:
For starters, apologies if a PR is the wrong way to start this conversation.

This is a bare bones POC chrome extension to populate login forms by connecting to a tab with keeweb loaded. What works now? When keyweb database is open, the extension can fill the login form for github (that's pretty much it).

I wanted to get some feedback, namely:
1. Is this something the user community is interested in getting? I'm planning to finish this for my personal use, but curious if others would find this useful.
2. Would this get integrated upstream in keeweb repo, or would it be best suited in a separate repo.
3. And I suppose building on question 2, the hacking I've done here to `app-model.js` is obviously not a final solution. I could see this going one of two ways:
3.1. Extension related code stays in keeweb core but gets abstracted into a module of some sort.
3.2. Extension related code is move to a plugin. I haven't fully wrapped my head around how plugins work so I'm curious to hear comments on this approach.

### DIFF
--- a/app/scripts/models/app-model.js
+++ b/app/scripts/models/app-model.js
@@ -811,6 +811,38 @@ class AppModel {
     }
 
     fileOpened(file, data, params) {
+        function findEntry(group, url) {
+            for (const entry of group.entries) {
+                if (entry.fields.URL === url) {
+                    return entry;
+                }
+            }
+
+            for (const g of group.groups) {
+                const entry = findEntry(g, url);
+                if (entry) {
+                    return entry;
+                }
+            }
+        }
+        window.postMessage({ type: 'file-opened' });
+        window.addEventListener('message', (event) => {
+            // only accept messages from the same frame
+            if (event.source !== window) {
+                return;
+            }
+
+            if (event.data.type === 'get-creds') {
+                const entry = findEntry(file.db.groups[0], event.data.url);
+                if (entry) {
+                    window.postMessage({
+                        type: 'creds',
+                        username: entry.fields.UserName,
+                        password: entry.fields.Password.getText()
+                    });
+                }
+            }
+        });
         if (file.storage === 'file') {
             Storage.file.watch(
                 file.path,
@@ -840,6 +872,7 @@ class AppModel {
     }
 
     fileClosed(file) {
+        window.postMessage({ type: 'file-closed' });
         if (file.storage === 'file') {
             Storage.file.unwatch(file.path);
         }
--- a/chrome/app-extension.js
+++ b/chrome/app-extension.js
@@ -1,0 +1,20 @@
+window.port = chrome.runtime.connect();
+window.port.onMessage.addListener(msg => {
+   window.postMessage(msg);
+});
+
+window.port.onDisconnect.addListener(() => {
+   console.log('app-extension.disconnect');
+   setTimeout(() => location.reload(), 1000);
+});
+
+window.addEventListener('message', event => {
+   console.log('app-extension.message', event);
+
+   // only accept messages from the same frame
+   if (event.source !== window) {
+      return;
+   }
+
+   window.port.postMessage(event.data);
+});
--- a/chrome/background.js
+++ b/chrome/background.js
@@ -1,0 +1,28 @@
+let keewebPort;
+
+chrome.runtime.onMessage.addListener(
+   function(request, sender, sendResponse) {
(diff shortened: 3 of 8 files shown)


antelle:
Hi! 🎁 
First of all, thank you for the contribution!

I agree, it would be more visible in issues, because PRs don't usually live long in open state and I try to either merge or reject them as soon as possible.
Anyways, let me share my thoughts on the future of browser auto-fill in KeeWeb.

The most common use case is KeeWeb desktop ↔ browser extension, this will be implemented using KeePassXC-browser protocol. KeeWeb desktop will expose API through a proxy and extensions will connect to it. This way we'll support existing browser extensions, and additionally there will be our own extension providing a better replacement for auto-type for browsers. All this is discussed in detail in #874.

Now, the interesting parts of it. Safari does not and will not support native messaging. And ~~I like Safari~~ I think it's a good idea to provide an extension for Safari as well, because they have WebExtensions support and it's not that hard. So, our own extension will also communicate using another transport, apart from Native Messaging. It will use the same KeePassXC-browser protocol on top of that, so that changes for Safari will be minimal, only on the transport level, which seems to be easy to achieve.

This all will however work only if you have a desktop app, which is not always possible. For example, on Chrome OS there's no such luck. That's here your scenario would work: KeeWeb is open in a browser tab (or a so-called Chrome "app"), and the extension connects to it.

I have a couple of questions here:

1. Protocol: obviously, there's no point in serialization when we have access to the API directly. The implementation should, however, use the same backend as will be implemented for KeePassXC-browser. We'll just not use the serialization part of it.
2. Communication: as it's noted before, we won't communicate over Native Messaging or TCP, so we'll use the API directly. Would we however expose it for everyone? Not sure, see the next item.
3. Security: since the extension will have access to KeeWeb tab, it's kind of meh. It would be ridiculous to add any questions because a malicious extension can just listen to password entry or override something in prototypes. Shall we then assume all open databases are enabled for auto-type in this way of integration? Probably yes. How to expose the API though? Ideally I would like to avoid having a global variable that will give access to all passwords, so that even it's possible, it's not that easy. I need to research communication between an extension and a browser tab to see what options are available, there should be some sort of extension identity verification.

How it will be organized in terms of repositories:

The core product (KeeWeb desktop and web) will include:
- implementation of KeePassXC-browser protocol logic
- serialization
- transport: Native Messaging and TCP for Safari
- security measures to limit access to this (question in the desktop app, see [here](https://github.com/keeweb/keeweb/wiki/Browser%20AutoFill#keewebhttp))
- enrollment, settings, links to browser extensions

KeeWebHttp will be sunset and removed from the plugin repository.

Extensions:
- there will be an official browser extension that will provide basic auto-fill, it will connect either to desktop app, or a tab, this extension will live in a separate repository in KeeWeb organization, something like `keeweb-browser`
- we'll of course provide support for other extensions implementing KeePassXC-browser
- you can develop 3rd-party extensions that will use the API provided by KeeWeb

Now, how to organize all of this now. Most likely, while we don't have any API, you can develop a KeeWeb plugin that will expose this API for your extension. For example, you can transform [KeeWebHttp](https://github.com/keeweb/keeweb-plugins/blob/master/docs/plugins/keewebhttp/plugin.js) to match your needs. Then, when the official API is implemented in KeeWeb, you will switch from your implementation to the one provided by KeeWeb. To simplify this switch, please follow the [KeePassXC-browser](https://github.com/keepassxreboot/keepassxc-browser/blob/develop/keepassxc-protocol.md) protocol.

antelle:
Ok, I did some research and I see that the question about allowing connection of course makes sense, while the communication with extensions will be via `postMessage`. This way other extensions won't get access to KeeWeb. So, the difference in your case will be that:
- the extension will use messages instead of other ways of communication
- it's a good question how to show entry selection and other dialogs, I'm not sure yet

antelle:
I started working on the extension [here](https://github.com/keeweb/keeweb-browser), but it will take time to build. Now it can't do much apart, but I hope to make something useful soon.

antelle:
Just as an update, the extension is complete: https://github.com/keeweb/keeweb-connect

jiggak:
Very cool! Does it work with the current keeweb @ https://app.keeweb.info ?

antelle:
No, for now only with bera.keeweb.info, I’m actively fixing bugs now, it needs some time to be production-ready

Issue #1904, opened 2021-09-24
CSV import always ignores last entry


howdytom:
**Describe the bug**
When I try to import 10 entries from a comma separated CSV file, KeeWeb always skips and ignores the last entry. This issue is serious if you don't want to miss any entries and check imported entries carefully!

**To Reproduce**
1. I have attached a test CSV file with 10 Login entries
2. Import [CSV_test-import-10entries.csv](https://github.com/keeweb/keeweb/files/7228706/CSV_test-import-10entries.csv) into KeeWeb 
3. KeeWeb imports Login 1-9
4. Login 10 will be ignored

**Expected behavior**
All entries inluding the last line should be imported.

**Workaround**
Create a additional helper entry which will be skipped.

**Kdbx File**
This issue occurs on a new and and existing Kdbx database.

antelle:
Hi! That's strange, thank you for spotting this!

HarlemSquirrel:
Fixed by #1944

PR #1944, opened 2021-12-05
Fixed issue with Csv parser parse (#1904)

R3dIO:

I have fixed the issue with CSV import ignoring the last entry as mentioned in issue (#1904). This bug was there because the parse method in csv-parser method was using (<) condition instead of (<=) that's why the last line is always ignored. So I fixed it by updating the check condition.


HarlemSquirrel (approved):
Thank you!

PR #1960, opened 2021-12-29
Use Wayland backend for Wayland sessions

GrantMoyer:
These two flags tell Electron, through Chromium, to use Wayland for rendering. The flags are set in the bash script, because setting them in desktop/main.js causes issues with rendering.

Rendering with Wayland instead of X11 and XWayland gives better support for DPI scaling. Additionally, this change also has implications for drag-and-drop, since drag-and-drop does not work consistently from XWayland to Wayland apps and visa versa (for example, from XWayland KeeWeb to Wayland Firefox under sway).

### DIFF
--- a/package/deb/usr/bin/keeweb
+++ b/package/deb/usr/bin/keeweb
@@ -1,2 +1,6 @@
 #!/bin/bash
-/usr/share/keeweb-desktop/keeweb "$@"
+if [ "$XDG_SESSION_TYPE" = "wayland" ]; then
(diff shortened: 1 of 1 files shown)


antelle:
Thank you! 🎁 
didn't know about this

GrantMoyer:
FYI, an upstream issue popped up since I submitted this, so it it's probably a good idea to wait on creating a release branch with this change.

https://github.com/electron/electron/issues/32487

Issue #1964, opened 2022-01-02
Task "eslint:desktop" failed


bmatusiak:
**Describe the bug**
following instructions in readme,  fails to build

**Expected behavior**
a successful build

**Environment**
linux   node-v12

**Logs**
```

/home/ec2-user/environment/brad/keeweb/keeweb/keeweb/desktop/scripts/util/browser-extension-installer.js
  18:93  error  Delete `,`                                                                                                                                                                                               prettier/prettier
  21:29  error  Replace `⏎························`${home}/Library/Application·Support/Mozilla/NativeMessagingHosts/`,⏎····················` with ``${home}/Library/Application·Support/Mozilla/NativeMessagingHosts/``  prettier/prettier
  26:99  error  Delete `,`                                                                                                                                                                                               prettier/prettier
  38:73  error  Delete `,`                                                                                                                                                                                               prettier/prettier
  41:29  error  Replace `⏎························`${home}/.mozilla/native-messaging-hosts/`,⏎····················` with ``${home}/.mozilla/native-messaging-hosts/``                                                    prettier/prettier
  45:29  error  Replace `⏎························`${home}/.config/microsoft-edge/NativeMessagingHosts/`,⏎····················` with ``${home}/.config/microsoft-edge/NativeMessagingHosts/``                            prettier/prettier
  59:67  error  Delete `,`                                                                                                                                                                                               prettier/prettier
  62:21  error  Replace `⏎················'HKCU\\Software\\Mozilla\\NativeMessagingHosts\\',⏎············` with `'HKCU\\Software\\Mozilla\\NativeMessagingHosts\\'`                                                      prettier/prettier
  66:21  error  Replace `⏎················'HKCU\\Software\\Microsoft\\Edge\\NativeMessagingHosts\\',⏎············` with `'HKCU\\Software\\Microsoft\\Edge\\NativeMessagingHosts\\'`                                      prettier/prettier

✖ 9 problems (9 errors, 0 warnings)
  9 errors and 0 warnings potentially fixable with the `--fix` option.

```

eliotmasset:
same for me

antelle:
Hi! You can try installing npm packages from a lockfile, maybe it will help.

PR #1987, opened 2022-02-17
Fix prettier (#1964)

qeeqez:
Quick fix for eslint\prettier error (#1964) during desktop build on developer branch.

No new functionality, only code styling issues.


antelle:
Thank you 🎁

PR #1998, opened 2022-03-17
Fixed missing localization of no title

ragnarstolsmark:
Fixed a couple of spots missing localization when an entry had no title.


Aetherinox:
These have been introduced in KeeWeb v1.19.0

PR #2002, opened 2022-03-26
Parameterize generator view options

GrantMoyer:
Before this change, generator view options were static, so all generators needed to work with the same options. This change parameterizes the options, so that plugins can provide different sets of options for custom generators. In particular, this change enables hiding the default options for generators where it doesn't make sense, such as the passphrase generator in keeweb/keeweb-plugins#30.

Example use:

parameterized_options

### DIFF
--- a/app/scripts/comp/app/generator-presets.js
+++ b/app/scripts/comp/app/generator-presets.js
@@ -2,11 +2,52 @@ import { AppSettingsModel } from 'models/app-settings-model';
 import { Locale } from 'util/locale';
 
 const GeneratorPresets = {
+    getBasicOptions() {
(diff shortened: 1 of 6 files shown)


Aetherinox:
Hey, the previous `develop` branch needed to be fixed, and it was best to create and utilize a new structure. If you wish to suggest / implement this change, please do a quick pull from the `develop/1.19.0` branch with your new changes, and then I'll review it.

If you wish to see if this is still something that should be implemented, take a look at the new branch and review the updated code. The password generator functionality has had many changes made to it, so we'd need to make sure these don't conflict with one another.

PR #2003, opened 2022-03-27
refactor: replace deprecated String.prototype.substr()

CommanderRoot:
[String.prototype.substr()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr) is deprecated so we replace it with [String.prototype.slice()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/slice) which works similarily but isn't deprecated.
.substr() probably isn't going away anytime soon but the change is trivial so it doesn't hurt to do it.


Aetherinox:
Thanks for this PR. I did a lot of deprecation work in 1.19. Sorry this PR took so long to address. We appreciate the contribution.
I'll merge this with the dev branch anyway so you can credit as contributing.

PR #2029, opened 2022-07-18
fix #2027: show space characters on reveal

### LINKED ISSUE #2027: Whitespace is not shown when revealing password
If you create a password which contains whitespace...
image

When you reveal it...
image

The whitespace is removed...
image

Editing the password shows the whitespace still...
image

And copying the password...
image

Includes the whitespace as well...
`whitespace between words`

KeeWeb v1.18.7 (dfc6201, 2021-07-18)
Environment: electron v12.0.7
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.18.7 Chrome/89.0.4389.128 Electron/12.0.7 Safari/537.36

qeeqez:
When obfuscating "revealed" password - we generate html like this:
`<div style="order: 3;"> </div>`

In this case SPACE is not shown. My idea is to replace SPACE with non-breaking space. It will generate such html:
`<div style="order: 3;">&nbsp;</div>`

### DIFF
--- a/app/scripts/util/formatting/password-presenter.js
+++ b/app/scripts/util/formatting/password-presenter.js
@@ -8,6 +8,8 @@ class RandomNameGenerator {
 }
 
 function charCodeToHtml(char) {
+    // In case of SPACE character we change to non-breaking space, so it will be shown as &#nbsp;
+    if (char === 32) char = 160;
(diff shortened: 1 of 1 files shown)


Aetherinox:
Thanks for providing this. Our recent series of updates have implemented a variation of this fix temporarily while we work on re-working the OTP feature.

Issue #1924, opened 2021-10-25
Changing to do not use a key file seems not working


PerezRE:
**Describe the bug**
After setup a key file, I decided to don't use one so I select 'Don't use a key file' option from settings, but after syncing or saving the changes, the same option changes itself to  'Use key file {filename}'. So can it be changed? or once the key file is setup, it can 't be removed?

**To Reproduce**
You need to have a keeweb with master password and key file already setup. Where keeweb file is sync in cloud storage service and the key file is locally stored.

1. Open your keeweb file
2. Go to setting's file
3. Select 'Dont use a key file' option from Key file.
4. Try to save or sync.
5. The error occurs reestablishing the option to 'Use key file exampleKefileName'

**Expected behavior**
it does supposed to allow to sign in with out a key file but still requiring the key file.

**Environment**
KeeWeb version: 1.18.7
OS: Windows 10 21H1 Pro x64.
Cloud Storage: OneDrive

gynet:
having the same issue

jamesdu031:
Same issue too

gmuenster:
Same issue too

rehberbey:
I think I found a similar error of the same error.

If you create a new key file but do not save it to your device, you must use that file (which does not exist) before you can log in.

![image](https://user-images.githubusercontent.com/58523422/223542058-6561113b-b936-49f7-8159-f88de4cd6c7e.png)

AlBundy33:
you can try to apply #2030 by yourself or you can also build a docker image with the pull request
https://github.com/AlBundy33/keeweb_kdbx4_hotfix

PR #2030, opened 2022-07-18
Fix remove key file

dissipative:
This fixes the issue #1924

So as kdbxweb has protected property `keyFileHash` in credentials, it must not been set to null without setter.

### DIFF
--- a/app/scripts/models/file-model.js
+++ b/app/scripts/models/file-model.js
@@ -536,7 +536,7 @@ class FileModel extends Model {
     }
 
     removeKeyFile() {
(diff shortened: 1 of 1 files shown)


Aetherinox:
@dissipative We'll check this out with priority. Need to run through some tests and ensure everything is functioning properly. This project recently added new developers; so we can start addressing these issues now.

Aetherinox:
Tested suggested change(s). No conflicts found.

Corrects the issue with being unable to unset a keyfile once you have created a database with a keyfile assigned.

Issue #1975, opened 2022-01-27
Crash on launch: GPU process isn't usable


neversun:
**Describe the bug**
App crashes on launch without showing any graphical interface

**To Reproduce**
```bash
➜  ~ keeweb                                         
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
libva error: vaGetDriverNameByIndex() failed with unknown libva error, driver_name = (null)
[39809:0127/140651.852037:FATAL:gpu_data_manager_impl_private.cc(415)] GPU process isn't usable. Goodbye.
[1]    39809 trace trap (core dumped)  keeweb
```

**Expected behavior**
App starts

**Environment**
KeeWeb v1.18.7 (dfc6201, 2021-07-18)
Environment: electron v12.0.7
User-Agent: Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) KeeWeb/1.18.7 Chrome/89.0.4389.128 Electron/12.0.7 Safari/537.36

Linux fedora 5.15.15-200.fc35.x86_64

**Notes**

This seems to be fixed starting with electron v13.5.0 (https://github.com/electron/electron/pull/31091)

Workaround by adding `--no-sandbox` to launch options: `keeweb --no-sandbox`
On Fedora 35, you can edit the desktop launch entry at `/usr/share/applications/keeweb.desktop`

antelle:
Well, adding `--no-sandbox` by default is probably not best, but we can update Electron, maybe it will fix that.

keyxmakerx:
This seems to happen on POP OS 22.04 as well now.

rutonyru:
Hi,
I'm playing with a lot of virtual machines and i migrate some to ubuntu 22.04. Adding --no-sandbox saved me ! 
Tips -- As I launch keeweb from favorite bar (ubuntu), i added the parameters to the `/usr/share/applications/keeweb.desktop` file.
Have a good day ;)
Antoine

AlexWayfer:
Now it fails on stable Arch Linux: https://bugs.archlinux.org/task/75902

Due to `glibc` minor update.

Please, update the Electron. And make a new release.

HarlemSquirrel:
For what it's worth, I switched to the appimage with my ArchLinux install and it's working again for me

ckruger097:
[...]

PR #2051, opened 2022-09-21
bump electron to 20

HarlemSquirrel:
I was able to load my db in my browser but `npm run electron` is showing a blank window so not sure what's going on there.

Addresses #1975

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "requires": true,
   "packages": {
     "": {
-      "version": "1.18.5",
+      "version": "1.18.7",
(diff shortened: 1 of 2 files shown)


Aetherinox:
A side note just in case anyone reviews this.

The reason that any bump above Electron 14 would not work is because all versions of Electron after v14 removes the built-in `electron.remote` API and migrates it to a new npm package [@electron/remote](https://www.npmjs.com/package/@electron/remote).

Currently have a bump planned for Electron after this section of KeeWeb is re-written.

PR #2074, opened 2022-12-18
bugfix(scripts): import otp error handling

ali4heydari:
Importing OTP with `otpauth://` that has `algorithm` and `type` in different case caused error


Aetherinox:
Thanks for this. These issues have been addressed in v1.19.0. I'll merge this with dev just so you get credit as contributing.

PR #2111, opened 2023-07-11
upgrade dependend packages for keeweb

ojehle:
Only updated the package.json with the actual versions not breaking the build.

-> could not build it with  @electon/remote and go to higher version then electron 13 latest release. the webpack did not pack  the package and it was missing in the result bundle.


carlbennett (approved):

Aetherinox:
Thanks for submitting this. Our recent work on updating all dependencies took care of this particular task.

The reason for Electron not bumping any higher than v13 is due to the fact that we make use of Electron.remote; which was removed from Electro v14 and made to be its own separate package which requires integration.

PR #2142, opened 2024-04-04
Add verify workflow

HarlemSquirrel:
Add new verify GitHub workflow that will run tests on pull requests and pushes.


HarlemSquirrel:
Well tests appear to be running but there's no progress indicator 🙁

HarlemSquirrel:
Tests seem to hang forever. Locally, if I cancel with `ctrl + c` after a while I see the progress. May need an older version of Node or to bump some dependency.

Aetherinox:
Yeah, right now the tests are hanging, ongoing for the last 28 minutes. I'm trying a few edits out on my other repo. The first initial test seemed to complete right away.

HarlemSquirrel:
No success so far.

I tried these node versions locally:
- v14.21.3
- v16.1.0
- v16.10.0
- v16.20.2

I also tried bumping grunt to 1.6.1 but didn't seem to make a difference.

Aetherinox:
> No success so far.
> 
> I tried these node versions locally:
> 
>     * v14.21.3
> 
>     * v16.1.0
> 
>     * v16.10.0
> 
>     * v16.20.2
> 
> 
> I also tried bumping grunt to 1.6.1 but didn't seem to make a difference.

Same with me. I've tried adding OpenSSL legacy support because sometimes it will pop off with an envelop error, which cleared the error because we're on an older version, but still didn't complete the tests.

I did however, remove the testing method before it starts the process, and it went through that time. So for whatever reason, the start of the tests are what is freezing the workflow runner.

Aetherinox:
The process halts at the line:
https://github.com/keeweb/keeweb/blob/e6df216291b9b5322049a11fe1301a2089679483/build/tasks/grunt-run-test.js#L18

It only runs up to the lines before.

antelle:
The build system gets outdated pretty fast, I assume now after two years there's a lot to fix there unfortunately.

Aetherinox:
I applied a fix and tested it on my Github repo via actions; and it works now.

```
    ✓ should generate a password with ambiguous characters
    ✓ should generate a password with custom characters
    ✓ should generate a password with special characters
    ✓ should generate a password with high ascii characters
    ✓ should generate a pronounceable password
    ✓ should generate a password with pattern
    ✓ should include all groups of characters at least once
  84 passing (79ms)
14
All tests passed
15
16
Done.
Execution Time (2024-04-04 17:52:16 UTC-0)
loading tasks     880ms  ▇▇▇▇▇▇▇▇ 17%
webpack:test       3.1s  ▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇▇ 60%
run-test:default   1.2s  ▇▇▇▇▇▇▇▇▇▇▇ 23%
Total 5.2s
```

I need to look through the code and make sure it didn't cause any other issues.

Aetherinox (approved):
Workflow properly executes after puppeteer argument specified.

Since you have access, I'll let you decide when to approve the changes. Not sure if you have anything else you want to do to it or test some more. I had a test one going but it didn't have anything extra other than just my own styling, and I didn't want to take your PR so I removed it.

.github/workflows/verify.yaml
```diff
@@ -0,0 +1,30 @@
+name: Verify
+on:
+  push:
+  pull_request:
+
```
Aetherinox, line 5: Might be good to add `workflow_dispatch:` just to test with.

Aetherinox:
Nice, it built that time. Progress.

PR #2174, opened 2024-05-21
Update to new UUID for firefox extension

HarlemSquirrel:
We are changing the UUID since we don't have access to the closed account or now revoked add-on so we need to use a new UUID for native messaging.

See https://github.com/keeweb/keeweb-connect/pull/55


Aetherinox:
Are you all done with this?

HarlemSquirrel:
> Are you all done with this?

Yup, it's ready to go.

PR #2179, opened 2024-06-15
Simplify font family and set `lang` attr to <html>

edo-bari-ikutsu:
In current version, fonts of Simplified Chinese are contained to `$base-font-family`, therefore Japanese and Traditional Chinese is displayed by using Simplified Chinese font. This is not good behavior for the aspect of i18n.

I propose to remove Simplified Chinese from the font setting, and instead of it, to set `lang` attribute to `html` element. By `lang` attribute, browser can determine which kanji font should be used.

I will show the examples as follows: 

[English]
English

[Japanese]
Japanese

[Simplified Chinese]
Simplified Chinese

[Traditional Chinese]
Traditional Chinese

Note: `$base-font-family: Arial, sans-serif;` setting is based on the settings of Google web page (https://www.google.com).

### DIFF
--- a/app/index.html
+++ b/app/index.html
@@ -1,6 +1,6 @@
 <!DOCTYPE html>
 <html>
-    <head lang="en">
+    <head>
         <meta charset="UTF-8" />
         <title>KeeWeb</title>
         <meta
--- a/app/scripts/app.js
+++ b/app/scripts/app.js
@@ -133,6 +133,7 @@ ready(() => {
                 }
             })
             .then(() => {
+                $('html').attr('lang', SettingsManager.activeLocale);
                 StartProfiler.milestone('loading remote config');
             });
     }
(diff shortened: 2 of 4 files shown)


antelle:
Looks very bad aesthetically in English. Perhaps in Chinese and Japanese it’s better, but not in Latin alphabets.

edo-bari-ikutsu:
@antelle Thank you for your feedback. I will try to add some Latin fonts to improve.

antelle:
I would suggest not changing the font-family for Latin fonts.

edo-bari-ikutsu:
I have restored the settings of latin fonts, bad aesthetically of latin character display has been solved.

Althouth non-latin fonts including Simplified Chinese are not included in `$base-font-family`, browser selects appropriate font by referring `lang` attr of `<html>`.

[English]
image

[Japanese]
image

[Simplified Chinese]
image

[Traditional Chinese]
image

codeCraft-Ritik (commented):
Great improvement! Using the lang attribute for font selection is a better approach for proper i18n handling.
This should ensure correct rendering across Japanese, Simplified, and Traditional Chinese 👍

Issue #322, opened 2016-08-20
WebDAV digest auth support


larsvik:
Hello,

Keeweb doesn't seem to work with remote (webdav) hosted databases. It only returns a 401 (unable to login/wrong u/p). The WebDAV server is an Apache2 with DAV mod and SSL/HTTPS. The server works with all WebDAV clients I've tried, including Keepass2 and Kypass 3.

BR,

LV

antelle:
Hello, 
Is your OPTIONS request [returning 200](https://github.com/keeweb/keeweb/wiki/WebDAV-Config) **without** authorization?
If yes, please show Network panel from devtools.

larsvik:
> AH01781: client used wrong authentication scheme `Basic': /wd/test.kdbx

OK, here's the problem. It seems Keeweb use basic auth, and not **digest** auth, which is more secure and is in fact a requirement by Windows 7 and later for WebDAV to work.

antelle:
Yes, only basic auth is supported for now

larsvik:
OK, great thank! :+1:

larsvik:
Oops, clicked the wrong button...

Tejes:
As a workaround, if you self-host KeeWeb on the same storage as the kdbx file, providing an index.html-relative link, and leaving user and password field empty will work fine. The browser takes care of WebDAV headers.

MetallianFR68:
Any update on this very old issue ?
It still prevents my user to store their kdbx file on their webdav server
Digest Authentication is needed

jiggak:
I'm working on implementing Digest Authentication (pull request coming soon).
Question regarding a design choice:
Should the option to pick between Digest/Basic authentication be in the app settings? Or should this just be inline with the web dav form where the user enters the URL, username, and password?

PR #2184, opened 2024-07-04
Webdav digest authentication

jiggak:
This PR aims to address the lack of support for digest authentication for webdav #322 

First, I should mention that this adds a dependency on https://www.npmjs.com/package/webdav.
I'm hoping an exception can be made to "don't add any dependencies" since this library provides a nice clean, robust webdav client implementation that greatly simplifies the webdav handling.

There are a few more things I need to do before I can call this ready to merge, but I'd like to get feedback/answers to a few questions.

### Translations
I added a setting for choosing between basic/digest authentication. ~How should I populate the French and German strings for the user facing messages?~ I used [LibreTranslate](https://libretranslate.com/) to fill in the French and German setting label, and left the "Basic" and "Digest" options as English since this seems to be the industry standard for technical terms.

### Alternative to `Last-Modified` header
Currently KeeWeb has a setting labelled "Always reload the file instead of relying on Last-Modified HTTP header". This setting causes the app to fetch the password file content instead of performing a more lightweight request (i.e. "stat") that gets file details, without fetching file content. This appears to have been added for instances where the webdav server doesn't return a `Last-Modified` header.

I think the implementation could be simplified to use `Etag` instead of `Last-Modified` as the file revision. In my testing `Etag` is available both in the response body of `PROPFIND` request, and as a response header in `GET` requests, so it would be ideal.
Assuming we can trust that all webdav servers will provide `Etag`, the "Always reload the file instead of relying on Last-Modified HTTP header" could be removed.

However, I'm happy to leave it... I will just need to re-implement the feature since I skipped that while refactoring.


falseUtopia:
Hi Hero, this pr how's this going? I need it 🤣, hope to merge it soon, thank you !

jiggak:
@Aetherinox  Is there anything I can do to get this change merged? It's a few months old, so I should/could get it brought up to current master.

Aetherinox:
Pull `develop/1.19.0` and add your changes there, then add a new PR I can go through. The repo did a lot of cleanup, and the develop branch was all over the place.

The develop/1.19.0 code is what is going to master when it's finished.

Aetherinox:
Just an update. If you're wanting this in 1.19.0, I'll need the merge soon. I'm doing bug fixes right now from users who have been testing v1.19.0 and adding in some last minute things. 

After 1.19.0, I don't know if there will be another minor release unless it's a bug fix, and we're going to migrate to v2.0 which will be framework changes.

jiggak:
Understood. I will get this rebased on 1.19.0 right now, and start a new PR.

Aetherinox:
Sounds good. Make sure you get the right branch. master is way behind. Right now master won't even build properly because of the outdated dependencies. That's the reason it has taken so long to get a new version out is we needed a stable build before we could do anything else. Now that it's stable, we can do whatever.

jiggak:
Yup, I'm rebasing on `develop/1.19.0`. I'll run the linter again to ensure it still passes, update the `webdav` dep as you suggested, and finally do some basic sanity tests.

Aetherinox:
Sounds good. If it's submitted after I'm done with bug fixes, I'll pull and test it out.

jiggak:
Created new PR #2218

Aetherinox:
Awesome. I'm finishing up a few things in the code, then I'll take a look.

PR #2218, opened 2024-12-21
Webdav digest authentication

jiggak:
This PR aims to address the lack of support for digest authentication for webdav #322 

First, I should mention that this adds a dependency on https://www.npmjs.com/package/webdav.
I'm hoping an exception can be made to "don't add any dependencies" since this library provides a nice clean, robust webdav client implementation that greatly simplifies the webdav handling.

There are a few more things I need to do before I can call this ready to merge, but I'd like to get feedback/answers to a few questions.

### Translations
I added a setting for choosing between basic/digest authentication. ~How should I populate the French and German strings for the user facing messages?~ I used [LibreTranslate](https://libretranslate.com/) to fill in the French and German setting label, and left the "Basic" and "Digest" options as English since this seems to be the industry standard for technical terms.

### Alternative to `Last-Modified` header
Currently KeeWeb has a setting labelled "Always reload the file instead of relying on Last-Modified HTTP header". This setting causes the app to fetch the password file content instead of performing a more lightweight request (i.e. "stat") that gets file details, without fetching file content. This appears to have been added for instances where the webdav server doesn't return a `Last-Modified` header.

I think the implementation could be simplified to use `Etag` instead of `Last-Modified` as the file revision. In my testing `Etag` is available both in the response body of `PROPFIND` request, and as a response header in `GET` requests, so it would be ideal.
Assuming we can trust that all webdav servers will provide `Etag`, the "Always reload the file instead of relying on Last-Modified HTTP header" could be removed.

However, I'm happy to leave it... I will just need to re-implement the feature since I skipped that while refactoring.


Aetherinox:
If you want to take a small dive, I need to look more in-depth tomorrow as I've been fixing bugs today.

But I noticed when I force it to fail due to incorrect headers, instead of getting an error message, it returns `err.response` with a null error value. But prior to the changes I got the standard `Network error`, or whatever else depending on the action triggered.

I'll look more in-depth tomorrow, if you want to poke around.

Also what setup did you use to test this. I used IIS and applied CORS headers through Traefik.

jiggak:
Taking a look at the issue with the error handling now.

Regarding how I've been testing... I've just been using my self hosted webdav service that runs on sabredav. I'll admit it's not the ideal thing to use since it's effectively my "production" instance.

jiggak:
Fixed the weird error message displayed when there was a CORS error, but now it's not that much better.

image

I was looking around at how error reporting is done elsewhere in the code. It seems like in most cases an `err` object is just passed directly through callbacks so I left it like that in the webdav impl. However, I'm happy to try and refine that if you'd like (specifically I'm sure I could get rid of the redundant `Error:` and `TypeError:` content in the messages).

Here is another example when CORS checks pass, but an authenticate error occurs.

image

Aetherinox:
> I was looking around at how error reporting is done elsewhere in the code. It seems like in most cases an err object is just passed directly through callbacks so I left it like that in the webdav impl. However, I'm happy to try and refine that if you'd like (specifically I'm sure I could get rid of the redundant Error: and TypeError: content in the messages).

Yeah I took a brief look last night just to see what was up, and I noticed that too.

That's entirely up to you if you wish to take a crack at it. I added "better errors" to the list on another roadmap, and I'm sure the users would greatly appreciate it if one of us took care of that. Because when I tried to force failures last night, the errors were very generic. The only time I knew exactly what was wrong is because I actually did something to cause it, or if I opened up dev console.

Aetherinox:
Let me know if you're up for re-doing the errors so I can allot some time (if not, that's fine). I just fixed a bunch of stuff, so I'm going to spend time on your PR tonight.

After I test it, I also want to bump the packages for kdbxweb and see if it still works. There's one package in that library that desperately needs updated due to an old vulnerability. And I want to focus on that right now and ensure it doesn't break your changes or anything else.

jiggak:
I would like to move this MR forward if possible. I suspect I will need to change the target branch for merge (rebase on master)?

PR #2259, opened 2025-09-22
Implement Google Picker for Drive file selection.

Nu11u5:
This pull request implements the Google Picker for allowing access to Google Drive files under the `drive.file` API scope.  Since the API scope is unrestricted the application does not require verification.

## User experience

When the user selects "Google Drive" to open a file, they will be directed to the OAuth consent flow.  After approving access the flow will take them to Google Picker, where they can select a KDBX file.

(At this point, the Drive service has granted the client permission to access the file.)

~~After selecting a file, the OAuth window will close and KeeWeb will list only the selected file.  The user must click the file to open it.~~

After selecting a file, the OAuth window will close and KeeWeb will automatically open the selected file.

After this, the UX continues normally.

### DIFF
--- a/app/scripts/const/cloud-storage-apps.js
+++ b/app/scripts/const/cloud-storage-apps.js
@@ -11,8 +11,8 @@ const DropboxApps = {
 
 const GDriveApps = {
     Local: {
-        id: '783608538594-36tkdh8iscrq8t8dq87gghubnhivhjp5.apps.googleusercontent.com',
-        secret: 'yAtyfc9TIQ9GyQgQmo3i0HAP'
+        id: '590412965107-i6g9vqm11ubkfiftimosti1ogddqit0e.apps.googleusercontent.com',
+        secret: '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'
     },
     Production: {
         id: '847548101761-koqkji474gp3i2gn3k5omipbfju7pbt1.apps.googleusercontent.com',
--- a/app/scripts/storage/impl/storage-dropbox.js
+++ b/app/scripts/storage/impl/storage-dropbox.js
@@ -245,7 +245,7 @@ class StorageDropbox extends StorageBase {
     }
 
     _apiCall(args) {
-        this._oauthAuthorize((err) => {
+        this._oauthAuthorize({}, (err) => {
             if (err) {
                 return args.error(err);
             }
--- a/app/scripts/storage/impl/storage-gdrive.js
+++ b/app/scripts/storage/impl/storage-gdrive.js
@@ -233,6 +233,7 @@ class StorageGDrive extends StorageBase {
                         const f = response;
                         const fileList = [
                             {
+                                auto: true,
                                 name: f.name,
                                 path: f.id,
                                 rev: f.headRevisionId,
--- a/app/scripts/storage/impl/storage-gdrive.js
+++ b/app/scripts/storage/impl/storage-gdrive.js
@@ -1,5 +1,4 @@
 import { StorageBase } from 'storage/storage-base';
-import { Locale } from 'util/locale';
 import { Features } from 'util/features';
 import { UrlFormat } from 'util/formatting/url-format';
 import { GDriveApps } from 'const/cloud-storage-apps';
@@ -50,7 +49,7 @@ class StorageGDrive extends StorageBase {
         if (path.lastIndexOf(NewFileIdPrefix, 0) === 0) {
             return callback && callback({ notFound: true });
         }
-        this._oauthAuthorize((err) => {
+        this._oauthAuthorize({}, (err) => {
             if (err) {
                 return callback && callback(err);
             }
@@ -78,7 +77,7 @@ class StorageGDrive extends StorageBase {
     }
 
     save(path, opts, data, callback, rev) {
-        this._oauthAuthorize((err) => {
+        this._oauthAuthorize({}, (err) => {
             if (err) {
                 return callback && callback(err);
             }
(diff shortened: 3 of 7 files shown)


Nu11u5:
This approach does have a few limitations:

- KeeWeb cannot see the directory structure on Drive, only a list of files that have been granted access.  The Picker UI is used to navigate the directory structure, and only the selected file is passed to the file list to keep it simple.
- Google Picker, when using the OAuth consent flow option, only allows filtering by MIME type, not file extension.  When a KDBX file is uploaded to Drive it is given the type `application/octet-stream` by default, which is not ideal in this case, since this applies to every unrecognized binary file type.  The Picker will filter on `application/octet-stream` as well as `application/x-keepass` and `x-application/kdbx` in case it ever used.
- The code is differentiating between calls to `StorageGDrive.list()` via the UI or other method by checking if `dir === undefined`.  This isn't great and will need to be kept consistent in the future, unless the code can be refactored to pass an event source or use a different method call.  I didn't look into modifying the UI code, only the storage.

Nu11u5:
Also, the code was pushed with my own Dev OAuth client credential.  For whatever reason the one provided failed.  I will be disabling the credential after this gets merged.

Nu11u5:
I've added an `auto` property to the file object and the UI will now skip displaying the file list and automatically open the first file passed with `auto == true`.

Aetherinox:
> KeeWeb cannot see the directory structure on Drive, only a list of files that have been granted access. The Picker UI is used to navigate the directory structure, and only the selected file is passed to the file list to keep it simple.

Yeah I noticed this.

I'm debating a new setting to switch between the two, the UI picker, and the original.

However, this fix comes at the same time where I've decided to integrate the API keys into the interface. However, because they're sensitive, I had to utilize the same functionality we use to secure passwords so that they're not sitting as raw text.

psociety:
Doesn't this break "Save to GDrive" feature?
image

Nu11u5:
> Doesn't this break "Save to GDrive" feature?

It shouldn't, since the `drive.file` scope allows an application to create new files with implicit ownership.  However, I will test this and create a fix if needed.

psociety:
> > Doesn't this break "Save to GDrive" feature?
> 
> It shouldn't, since the `drive.file` scope allows an application to create new files with implicit ownership. However, I will test this and create a fix if needed.

Just in case it helps, I will list the steps that make me get the error consistently:
1. I click New
2. I click New at the bottom to go to the settings page
3. I click Save As.. -> Google Drive.
BOoom explosion.

My workaround has been to make `list` method work with both the new and the old code:
```js
list(dir, callback) {
        const opts = {};
        if (dir === undefined) {
            opts.forceAuth = true;
            opts.urlParams = {
                'prompt': 'consent',
                'trigger_onepick': 'true',
                'mimetypes': 'application/x-keepass,x-application/kdbx,application/octet-stream'
            };
        }
        this._oauthAuthorize(opts, (err) => {
            if (err) {
                return callback && callback(err);
            }
            this.logger.debug('List');

            const ts = this.logger.ts();

            if (dir === 'drives') {
                const urlParams = {
                    pageSize: 100
                };

                const url = UrlFormat.makeUrl(`${this._baseUrl}/drives`, urlParams);

                this._xhr({
                    url,
                    responseType: 'json',
                    success: (response) => {
                        if (!response) {
                            this.logger.error('Drive list error', this.logger.ts(ts));
                            return callback?.('drive list error');
                        }
                        this.logger.debug('Listed drives', this.logger.ts(ts));

                        const fileList = response.drives.map((d) => ({
                            name: d.name,
                            path: d.id,
                            dir: true
                        }));
                        return callback?.(null, fileList);
                    },
                    error: (err) => {
                        this.logger.error('Drive dist error', this.logger.ts(ts), err);
                        return callback?.(err);
                    }
                });
            } else {
                let urlParams;
                if (this._pickedFileId) {
                    urlParams = {
                        fields: 'id,name,mimeType,headRevisionId',
                        supportsAllDrives: true
                    };
                } else {
                    let query = 'trashed=false and ';
                    if (dir === 'shared') {
                        query += 'sharedWithMe=true';
                    } else if (dir) {
                        query += `"${dir}" in parents`;
                    } else {
                        query += '"root" in parents';
                    }
                    urlParams = {
                        fields: 'files(id,name,mimeType,headRevisionId)',
                        q: query,
                        pageSize: 1000,
                        includeItemsFromAllDrives: true,
                        supportsAllDrives: true
                    };
                }

                const url = UrlFormat.makeUrl(
                    `${this._baseUrl}/files${this._pickedFileId ? `/${this._pickedFileId}` : ''}`,
                    urlParams
                );

                this._xhr({
                    url,
                    responseType: 'json',
                    success: (response) => {
                        if (!response) {
                            this.logger.error('List error', this.logger.ts(ts));
                            return callback?.('list error');
                        }
                        this.logger.debug('Listed', this.logger.ts(ts));
                        let fileList;
                        if (this._pickedFileId) {
                            const f = response;
                            fileList = [
                                {
                                    auto: true,
                                    name: f.name,
                                    path: f.id,
                                    rev: f.headRevisionId,
                                    dir: f.mimeType === 'application/vnd.google-apps.folder'
                                }
                            ];
                        } else {
                            fileList = response.files.map((f) => ({
                                name: f.name,
                                path: f.id,
                                rev: f.headRevisionId,
                                dir: f.mimeType === 'application/vnd.google-apps.folder'
                            }));
                            if (!dir) {
                                fileList.unshift({
                                    name: Locale.gdriveSharedWithMe,
                                    path: 'shared',
                                    rev: undefined,
                                    dir: true
                                });
                                fileList.unshift({
                                    name: Locale.gdriveSharedDrives,
                                    path: 'drives',
                                    rev: undefined,
                                    dir: true
                                });
                            }
                        }

                        return callback?.(null, fileList);
                    },
                    error: (err) => {
                        this.logger.error('List error', this.logger.ts(ts), err);
                        return callback?.(err);
                    }
                });
            }
        });
    }
```

Hope it helps!

codeCraft-Ritik (commented):
Great feature! Integrating Google Picker with a smooth OAuth flow significantly improves usability for Drive access.
The UX flow is clear and well thought out

codeCraft-Ritik (commented):
Scope Verification : Using drive.file is an excellent choice as it adheres to the principle of least privilege.

PR #2271, opened 2026-04-30
Prepare 1.18.9

HarlemSquirrel:


HarlemSquirrel:
@Aetherinox Just trying to get a build working again so this will work with latest Firefox addon that had to change the addon ID.

